PR #2, opened 2021-08-20
Basic Vite project

zachmargolis:
Super basic vite project

I ended up picking plain JS with TS annotations since I like the idea of being able to copy-paste into the browser easily, but I'm not married to it


README.md
```diff
@@ -2,3 +2,16 @@
 
 Summary reports of login.gov activity
 
+## Development
+
+### Run Development Server
+
+```bash
+npm install
+```
```
aduth, line 11: This won't run it on its own? Either needs a `postinstall` script in `package.json` (not preferred) or extra `npm run dev` command listed here.

package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "identity-reporting",
+  "version": "0.0.0",
+  "scripts": {
+    "dev": "vite",
+    "build": "vite build",
+    "serve": "vite preview",
+    "test": "run-p test:*",
+    "test:typecheck": "tsc",
+    "test:lint": "eslint src --ext .js",
```
aduth, line 10: `.js` is default, so I think it can be dropped.

```suggestion
    "test:lint": "eslint src",
```

src/report.js
```diff
@@ -0,0 +1,24 @@
...
+ * @typedef PathProps
+ * @prop {string} reportName
+ * @prop {Date} date
+ * @prop {string=} env 
+ */
+
+/**
+ * @param {PathProps} props
+ */
+function path({ reportName, date, env = 'prod' }) {
```
aduth, line 16: Unclear if this is meant to be used as a component. If yes, should be capitalized "Path". If no, "props" is a bit misleading of an argument name.

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+		"target": "ESNext",
+		"strict": true,
+    "noImplicitAny": true,
+		"noEmit": true,
```
aduth, line 8: Mixed tabbing.

```suggestion
    "target": "ESNext",
    "strict": true,
    "noImplicitAny": true,
    "noEmit": true,
```

.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
```
aduth, line 1: For future consideration: If Prettier's already in the project, may be good to enforce via [`eslint-plugin-prettier`](https://github.com/prettier/eslint-plugin-prettier).

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": true,
+		"target": "ESNext",
+		"strict": true,
+    "noImplicitAny": true,
```
aduth, line 7: `noImplicitAny` defaults to `true` when `strict` is set:

https://www.typescriptlang.org/tsconfig#noImplicitAny

package-lock.json
```diff
@@ -0,0 +1,2226 @@
+{
+  "name": "identity-reporting",
+  "version": "0.0.0",
+  "lockfileVersion": 1,
```
aduth, line 4: Up for consideration whether it's worth getting on NPM 7 now vs. later, which will reformat this as `"lockfileVersion": 2`. I can't recall if this is default version with current Node LTS, though LTS is also bumping to 16 in October.

package.json
```diff
@@ -0,0 +1,30 @@
...
+    "test:unit": "mocha -r esbuild-register 'src/**/*.test.js'"
+  },
+  "dependencies": {
+    "d3-time-format": "^4.0.0",
+    "htm": "^3.1.0"
```
aduth, line 15: Not in use currently. Fine to keep if you'd planned to use it in the future though.

aduth (approved):

.circleci/config.yml
```diff
@@ -0,0 +1,44 @@
...
+      - run:
+          name: Run tests
+          command: npm test
+
+workflows:
+  # Below is the definition of your workflow.
+  # Inside the workflow, you provide the jobs you want to run, e.g this workflow runs the build-and-test job above.
+  # CircleCI will run this workflow on every commit.
+  # For more details on extending your workflow, see the configuration docs: https://circleci.com/docs/2.0/configuration-reference/#workflows
+  sample:
```
aduth, line 40: Should we give it a real name?

README.md
zachmargolis: whoops, I meant to add `npm run dev` on the second line 🤦

src/report.js
zachmargolis: no this is just a utility function, I just though this was the syntax for annotating a a function's keyword arguments

package.json
zachmargolis: Good catch, removed in d11e28bf43daf06668fb4312f81b8e85f104c857

.circleci/config.yml
zachmargolis: Yup, got it in 2a1671fc541c6d4f66eb8e0639e664841135d647

src/report.js
aduth: > no this is just a utility function, I just though this was the syntax for annotating a a function's keyword arguments

Just a naming quibble. I might call the object "options" or something? At least if we plan to use React, "props" would make me think of a component.

tsconfig.json
zachmargolis: Removed in e1afcf0627b2414999063c1e83317e3b08b8a85e

src/report.js
zachmargolis: got it! updated in 5e8674b63b8f29ee2197ac1232b130a59200f51a

.eslintrc.json
zachmargolis: I intended to add it, fixed in ed35e669b37c35e8f34fc11bd7ee5622bbc75c40

package-lock.json
zachmargolis: 🤷 if you think it's worthwhile, happy to give that a shot? what do we have to change in the package.json to make that happen?

aduth: Just need to run `npm install` after upgrading. Which is be partly why it might be good to do now, since it tends to be confusing for someone already running npm 7 when it introduces local changes simply by installing 😬

zachmargolis: Should we also pin an node version in package.json?

aduth: > Should we also pin an node version in package.json?

I don't think `engines` will do much for us in terms of encouragement or enforcement. Doesn't hurt to include though. And maybe an `.nvmrc` file would be useful?

zachmargolis: Ok! Do you mind making a PR with a version you think would make sense?

aduth: Sure 👍  See #4.

PR #3, opened 2021-08-20
Set up project to build in Federalist

zachmargolis:
- output to _site

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,10 +5,9 @@
     "build": "vite build",
     "dev": "vite",
     "federalist": "npm run build -- --base=$BASEURL",
-    "prettier:fix": "prettier --write src/",
+    "prettier:fix": "npm run test:lint -- --fix",
     "serve": "vite preview",
-    "test:lint": "eslint src",
-    "test:prettier": "prettier --check src/",
+    "test:lint": "eslint .",
     "test:typecheck": "tsc",
     "test:unit": "mocha -r esbuild-register 'src/**/*.test.js'",
     "test": "run-p test:*"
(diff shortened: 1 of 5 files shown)


zachmargolis:
example works in federalist preview! https://federalist-14721e0a-0772-4dcd-b62f-ee9d20dcc5fd.app.cloud.gov/preview/18f/identity-reporting/margolis-federalist/

vite.config.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  build: {
+    outDir: '_site'
+  }
+}
```
aduth, line 5: Looks like Prettier isn't being run on this file, since `eslint` is only running against `src`. Could change `test:lint` script to run `eslint .`, though might need to ignore a few paths with an `.eslintignore`.

package.json
```diff
@@ -4,6 +4,7 @@
   "scripts": {
     "build": "vite build",
     "dev": "vite",
+    "federalist": "npm run build -- --base=$BASEURL",
     "prettier:fix": "prettier --write src/",
```
aduth, line 8: I didn't notice 'til now, but with Prettier integrated into ESLint, we shouldn't need a separate `test:prettier`, since that'll be included in the lint. Also, a `prettier:fix` could be implemented as `npm run test:lint -- --fix`.

aduth (approved):
🎉

package.json
zachmargolis: neat, updated in 3a390a4a42321ea192d31a3980665faf25a7f244

PR #4, opened 2021-08-23
Upgrade package-lock.json lockfileVersion, add nvmrc

aduth:
Context: https://github.com/18F/identity-reporting/pull/2#discussion_r693245681

**Why**: Avoid undesirable workspace changes for developers who may already be running npm@>=7

**Why**: For greater alignment to Federalist supports, and to support developers who may be using NVM with auto-switch shell integration.

**Additional information:**

- Lockfile version 2 is claimed to be backward-compatible for developers using npm@<7 ([source](https://github.blog/2021-02-02-npm-7-is-now-generally-available/)), so it's not strictly necessary to require version 7 or newer.

**Open questions:**

- As an addendum to my comment at https://github.com/18F/identity-reporting/pull/2#discussion_r693916992, we _could_ use `engines` as an enforcement measure in combination with [`.npmrc` `engine-strict`](https://docs.npmjs.com/cli/v6/using-npm/config#engine-strict), though this may be overkill since running older versions may be fine.
- `lts/*` is a moving target, but one intended to align to Federalist support. Alternatively, we could pin to a specific version like `14` (the current LTS until October).


aduth:
Interesting build failure:

>If you're using npm v7, make sure your package-lock.json file contains either
>"lockfileVersion": 1 or the code "hasInstallScript": true. If it doesn't have
>either of those, then it is likely the case that a known bug in npm v7 has
>corrupted your package-lock.json file. Regenerating your package-lock.json file
>should fix this issue.

Will try to regenerate.

zachmargolis (approved):
LGTM, thanks for adding! looks like something in circleci got mad, but assuming we get past that 👍

PR #11, opened 2021-08-27
Set ESLint warnings to errors

aduth:
**Why**: Because most people ignore warnings, so we should either enforce it or drop them from the configuration.

The approach was to add `error` entries for any in [the plugin's `recommended.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.ts) set to `warn`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,23 +1,27 @@
 {
   "root": true,
   "parser": "@typescript-eslint/parser",
+  "plugins": ["prettier", "@typescript-eslint"],
+  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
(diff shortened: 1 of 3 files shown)


aduth:
> > trying to add missing types for `@observablehq/plot` is daunting 😱
> 
> 😱 Can we just disable the no-explicit-any for the entire `plot` file?

Yeah, that seems like a reasonable short-term solution. I'd normally not mind adding the type definitions, but based on a quick glance of [the README](https://www.npmjs.com/package/@observablehq/plot), I don't think that's going to be very straight-forward 😢

zachmargolis (approved):
LGTM

PR #12, opened 2021-08-31
Use login.gov shared ESLint configuration

aduth:
Related: https://github.com/18F/identity-idp/pull/5359

**Why:** So that we can use a common standard ruleset across all repositories where JavaScript is used.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,6 +20,7 @@
     "@typescript-eslint/no-non-null-assertion": "error",
     "@typescript-eslint/no-unused-vars": "error",
     "import/no-extraneous-dependencies": ["error", { "devDependencies": ["**/*.test.*", "./*"] }],
+    "jsx-a11y/label-has-associated-control": ["error", { "assert": "either" }],
     "react/jsx-filename-extension": ["error", { "extensions": [".tsx"] }]
   },
   "settings": {
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,27 +1,32 @@
 {
   "root": true,
   "parser": "@typescript-eslint/parser",
-  "plugins": ["prettier", "@typescript-eslint"],
-  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended"],
+  "plugins": ["@18f/eslint-plugin-identity", "@typescript-eslint"],
+  "extends": [
+    "plugin:@18f/eslint-plugin-identity/recommended",
+    "plugin:@typescript-eslint/recommended"
+  ],
   "parserOptions": {
     "sourceType": "module",
     "ecmaVersion": 2021
   },
   "env": {
     "node": true,
-    "browser": true,
-    "mocha": true,
-    "es6": true
+    "browser": true
   },
   "rules": {
     "@typescript-eslint/explicit-module-boundary-types": "error",
     "@typescript-eslint/no-explicit-any": "error",
     "@typescript-eslint/no-non-null-assertion": "error",
     "@typescript-eslint/no-unused-vars": "error",
-    "prettier/prettier": "error",
-    "eqeqeq": "error",
-    "semi": "error",
-    "no-shadow": "error"
+    "react/jsx-filename-extension": ["error", { "extensions": [".tsx"] }]
+  },
+  "settings": {
+    "react": {
+      "pragma": "h",
+      "fragment": "Fragment",
+      "version": "16.0"
+    }
   },
   "ignorePatterns": ["_site/**"]
 }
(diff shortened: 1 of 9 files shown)


.eslintrc.json
```diff
@@ -1,27 +1,33 @@
...
   "rules": {
     "@typescript-eslint/explicit-module-boundary-types": "error",
     "@typescript-eslint/no-explicit-any": "error",
     "@typescript-eslint/no-non-null-assertion": "error",
     "@typescript-eslint/no-unused-vars": "error",
-    "prettier/prettier": "error",
-    "eqeqeq": "error",
-    "semi": "error",
-    "no-shadow": "error"
+    "quotes": ["error", "double"],
```
aduth, line 22: We should switch this in the Prettier config as a separate pull request.

zachmargolis:
```
identity-reporting/src/table.tsx
  12:3   error  propType "numberFormatter" is not required, but has no corresponding defaultProps declaration  react/require-default-props
```

Does that mean we just change this:
```diff
-  numberFormatter?: (n: number) => string;
+  numberFormatter: (n: number) => string;
```

zachmargolis (approved):
I tried checking this out locally but all I got was a bunch of `@typescript-eslint/no-explicit-any` warning errors?

aduth:
> I tried checking this out locally but all I got was a bunch of `@typescript-eslint/no-explicit-any` warnings, no errors?

Couple thoughts:

- Did you run `npm install`?
- Is `identity-idp` in the parent directory from this one?
- Is `identity-idp` checked out to the `aduth-eslint-plugin` branch?

I also had issues that it wasn't picking up dependencies defined in the package, so I had to manually run this as well:

```
npm i --no-save @babel/eslint-parser @babel/eslint-plugin eslint-config-airbnb eslint-plugin-import eslint-plugin-jsx-a11y eslint-plugin-mocha eslint-plugin-prettier eslint-plugin-react eslint-plugin-react-hooks
```

I think it's only a temporary issue for local testing. Surprised `yarn link` wasn't working great either.

aduth:
> Does that mean we just change this:

I'm thinking we just turn off the rule. `defaultProps` isn't something we should care about, since we're typically not using class components and can define default props with function arguments object destructuring defaults.

zachmargolis:
> Couple thoughts:
> 
> * Did you run `npm install`?
> * Is `identity-idp` in the parent directory from this one?
> * Is `identity-idp` checked out to the `aduth-eslint-plugin` branch?

Yes to all those

> I also had issues that it wasn't picking up dependencies defined in the package, so I had to manually run this as well:

No, I hadn't run this, will give it a shot

zachmargolis:
Ok got it running, the `comma-dangle` rule is triggering, but then prettier disagrees with it, in table.tsx:

```diff
        {body.map((row) => (
          <tr>
            {row.map((d) =>
              typeof d === "number" ? (
                <td className="table-number">{numberFormatter(d)}</td>
              ) : (
                <td>{d}</td>
-             )
+             ),
            )}
          </tr>
        ))}
```

aduth:
> Ok got it running, the `comma-dangle` rule is triggering, but then prettier disagrees with it, in table.tsx:

Hm, I thought these might have been because we set Prettier's [`"trailingComma": "es5"`](https://github.com/18F/identity-reporting/blob/c98e501c2a9f21401f783b697f89f86a7a6122d0/.prettierrc.json#L4), but it looks like changing that doesn't make a difference. And while I agree with the premise of the rule, it seems to be another stylistic thing we could just defer to Prettier. Technically this could apply to [`quotes`](https://eslint.org/docs/rules/quotes) as well, though I think there's more an argument there toward consistency regardless of how one has Prettier configured.

Also has me wondering if we can somehow bundle a "default" Prettier setup, at least to use in the absence of a config file.

package.json
```diff
@@ -21,6 +21,7 @@
     "@observablehq/plot": "^0.2.0",
     "d3-array": "^3.0.2",
     "d3-format": "^3.0.1",
+    "d3-time": "^3.0.0",
```
zachmargolis, line 24: ahah whoops, I bet this only worked because observable includes all of d3

aduth: > ahah whoops, I bet this only worked because observable includes all of d3

Yeah that sounds about right

src/daily-auths-report.tsx
```diff
@@ -36,18 +22,33 @@ interface Result {
    */
   iaa?: string;
 
+  // eslint-disable-next-line camelcase
   friendly_name: string;
```
aduth, line 26: We could maybe consider mapping this back to a camel case property as part of `process`. Didn't really want to deal with it here.

PR #14, opened 2021-09-01
Use binX because dates are continuous data

zachmargolis:
Using binning like this lets us take advantage of smarter axis labels

| Before | After |
| --- | -- |
| Screen Shot 2021-08-31 at 5 21 56 PM | Screen Shot 2021-08-31 at 5 21 36 PM |

### DIFF
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -2,7 +2,7 @@ import { path as reportPath } from "./report";
 import { VNode } from "preact";
 import { useContext, useEffect, useRef } from "preact/hooks";
 import { ReportFilterControlsContext } from "./report-filter-controls";
-import { utcDays } from "d3-time";
+import { utcDays, utcDay } from "d3-time";
 import * as Plot from "@observablehq/plot";
 import { format } from "d3-format";
 import { useQuery } from "preact-fetching";
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -1,14 +1,14 @@
+import { path as reportPath } from "./report";
 import { VNode } from "preact";
 import { useContext, useEffect, useRef } from "preact/hooks";
-import { utcDay, utcDays } from "d3-time";
+import { ReportFilterControlsContext } from "./report-filter-controls";
+import { utcDays } from "d3-time";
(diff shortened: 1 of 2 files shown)


src/daily-auths-report.tsx
```diff
@@ -145,18 +145,32 @@ function DailyAuthsReport(): VNode {
...
+            data || [],
+            Plot.binX(
+              {
+                y: "sum",
+              },
+              {
+                y: "count",
+                x: {
+                  value: "date",
+                  thresholds: utcDay,
```
aduth, line 163: Checking my understanding: According to [the documentation](https://www.npmjs.com/package/@observablehq/plot), this is passed...

>three arguments: the array of input values, the domain minimum, and the domain maximum.

Is that what we're expecting here?

Aside: I notice for shorter timespans, it can add a mid-day label. Are we okay with that?

image

aduth (approved):

src/daily-auths-report.tsx
zachmargolis: > Checking my understanding: According to [the documentation](https://www.npmjs.com/package/@observablehq/plot), this is passed...
> 
> > three arguments: the array of input values, the domain minimum, and the domain maximum.

Screen Shot 2021-09-01 at 8 36 10 AM

I guess the specific type isn't clearly, but I think that "a time interval (for temporal binning)" means an interval function like from d3-time



> Aside: I notice for shorter timespans, it can add a mid-day label. Are we okay with that?

Normally it would be just fine to let D3 auto-pick the interval, however, for shorter spans, it picks 12-hour interval and unforunately our data is only at midnight so we get these weird gaps:

Screen Shot 2021-09-01 at 8 40 15 AM

so I worked around that by pinning to `utcDay` specifically to avoid that. I think the labels might keep some of that same previous logic, and I think it's okay for now

aduth: >I guess the specific type isn't clearly, but I think that "a time interval (for temporal binning)" means an interval function like from d3-time

Ah, yeah, that was a bit unclear. Although, checking again, I think I assumed `utcDay` was a function, when in fact it's an object with `count` and `every`. Just a misread on my part, makes sense now 👍

PR #15, opened 2021-09-13
Adds basic identity-style-guide styles

zachmargolis:
Adding in our styles, cleaning up the form and the table a tad

| Before | After|
| --- | --- |
| Screen Shot 2021-09-13 at 10 18 35 AM | Screen Shot 2021-09-13 at 10 18 16 AM|
| Screen Shot 2021-09-13 at 10 18 42 AM | Screen Shot 2021-09-13 at 10 18 24 AM |

### DIFF
--- a/index.html
+++ b/index.html
@@ -93,10 +93,9 @@
         </div>
       </section>
     </header>
-    <main>
+    <main class="margin0 padding-1">
       <div id="app"></div>
     </main>
     <script type="module" src="/src/main.tsx"></script>
-    <script src="node_modules/uswds/dist/js/uswds.js"></script>
   </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -6,7 +6,97 @@
     <title>Login.gov Reporting Dashboard</title>
   </head>
   <body>
-    <div id="app"></div>
+    <header>
+    <section class="usa-banner" aria-label="Official government website">
+      <div class="usa-accordion">
+        <header class="usa-banner__header">
+          <div class="usa-banner__inner">
+            <div class="grid-col-auto">
+              <img
+                class="usa-banner__header-flag"
+                src="node_modules/identity-style-guide/dist/assets/img/us_flag_small.png"
+                alt="U.S. flag"
+              />
+            </div>
+            <div class="grid-col-fill tablet:grid-col-auto">
+              <p class="usa-banner__header-text">
+                An official website of the United States government
+              </p>
+              <p class="usa-banner__header-action" aria-hidden="true">Here’s how you know</p>
+            </div>
+            <button
+              class="usa-accordion__button usa-banner__button"
+              aria-expanded="false"
+              aria-controls="gov-banner-default"
+            >
+              <span class="usa-banner__button-text">Here’s how you know</span>
+            </button>
+          </div>
+        </header>
+        <div class="usa-banner__content usa-accordion__content" id="gov-banner-default">
+          <div class="grid-row grid-gap-lg">
+            <div class="usa-banner__guidance tablet:grid-col-6">
+              <img
+                class="usa-banner__icon usa-media-block__img"
+                src="node_modules/identity-style-guide/dist/assets/img/icon-dot-gov.svg"
+                role="img"
+                alt=""
+                aria-hidden="true"
+              />
+              <div class="usa-media-block__body">
+                <p>
+                  <strong> Official websites use .gov </strong>
+                  <br />
+                  A <strong>.gov</strong> website belongs to an official government organization in
+                  the United States.
+                </p>
+              </div>
+            </div>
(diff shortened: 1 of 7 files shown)


src/main.tsx
```diff
@@ -1,8 +1,16 @@
 import "./css/style.css";
+import "../node_modules/identity-style-guide/dist/assets/scss/_styles.scss";
+
 import { render } from "preact";
 import { Link } from "./router";
 import { Routes } from "./routes";
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import * as USWDS from "../node_modules/uswds/dist/js/uswds";
```
zachmargolis, line 9: This disable comment did not get rid of the error I wanted it to :/

```
src/main.tsx:9:24 - error TS7016: Could not find a declaration file for module '../node_modules/uswds/dist/js/uswds'. 'identity-reporting/node_modules/uswds/dist/js/uswds.js' implicitly has an 'any' type.

9 import * as USWDS from "../node_modules/uswds/dist/js/uswds";
```

zachmargolis: there's probably a better place to put this import, but adding the `<script>` in `index.html` didn't seem to work, might need to try again

src/main.tsx
```diff
@@ -1,8 +1,16 @@
...
+
+/* eslint-disable @typescript-eslint/no-unused-vars */
+const isUSWDSLoaded = typeof USWDS !== "undefined";
```
zachmargolis, line 12: the reason for this was so the import above didn't get pruned by vite

src/main.tsx
aduth, line 9: I'd not expect to see USWDS referenced directly here, since we use (and depend on) `identity-style-guide`.

Might help with your other issues, but could we do one or the other of...

```suggestion
import "identity-style-guide/auto";
```

Or:

```suggestion
import { banner /*, ... */ } from "identity-style-guide";
[ banner /*, ... */ ].forEach((component) => component.on());
```

Ref:

- https://github.com/18F/identity-idp/blob/14364e5/app/javascript/app/components/index.js
- https://github.com/18F/identity-style-guide#javascript-package

package.json
```diff
@@ -19,10 +19,12 @@
   },
   "dependencies": {
     "@observablehq/plot": "^0.2.0",
+    "autoprefixer": "^10.3.4",
     "d3-array": "^3.0.2",
     "d3-format": "^3.0.1",
     "d3-time": "^3.0.0",
     "d3-time-format": "^4.0.0",
+    "postcss-cli": "^8.3.1",
```
aduth, line 27: Do we need this?

src/report-filter-controls.tsx
```diff
@@ -81,18 +81,28 @@ function ReportFilterControls({
...
       <form ref={formRef} onChange={update}>
         <div>
           <label>
             Start
-            <input type="date" name="start" value={yearMonthDayFormat(start)} />
+            <input
+              type="date"
+              name="start"
+              value={yearMonthDayFormat(start)}
+              className="usa-input"
```
aduth, line 93: I don't know if we'd want to use it, especially since a lot of the JavaScript-heavy design system components don't work so well in React applications, but there is a design system date picker we could use:

https://designsystem.digital.gov/components/date-picker/

aduth (commented):
I don't see all of the styles from the original comment screenshot, specifically next / previous week buttons:

image

package.json
zachmargolis: probably not? I must have copy-pasted from one README or another... will remove

index.html
```diff
@@ -6,7 +6,96 @@
...
+                    Share sensitive information only on official, secure websites.
+                  </p>
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+      </section>
+    </header>
+    <main class="margin0 padding-1">
```
aduth, line 96:
```suggestion
    <main class="margin-0 padding-1">
```

src/report-filter-controls.tsx
zachmargolis: The native one in Chrome seems to be OK for now?
Screen Shot 2021-09-13 at 2 21 16 PM

index.html
```diff
@@ -6,7 +6,96 @@
...
+                    <br />
+                    A <strong>.gov</strong> website belongs to an official government organization in
+                    the United States.
+                  </p>
+                </div>
+              </div>
+              <div class="usa-banner__guidance tablet:grid-col-6">
+                <img
+                  class="usa-banner__icon usa-media-block__img"
+                  src="/node_modules/identity-style-guide//assets/img/icon-https.svg"
```
zachmargolis, line 58:
```suggestion
                  src="/node_modules/identity-style-guide/assets/img/icon-https.svg"
```

aduth (approved):
Looks good 👍

PR #16, opened 2021-09-14
Add ?env= param that lets us point the web at different backends

zachmargolis:
- For now, no corresponding UI param

open to feedback! maybe add it in a collapsible "debug" menu or something

Screen Shot 2021-09-14 at 11 08 58 AM

### DIFF
--- a/src/report.ts
+++ b/src/report.ts
@@ -3,18 +3,31 @@ import { utcFormat } from "d3-time-format";
 const yearFormat = utcFormat("%Y");
 const yearMonthDayFormat = utcFormat("%Y-%m-%d");
 
+function domain(env: string): string {
+  switch (env) {
+    case "local":
+      return "";
+    case "prod":
+    case "staging":
+    case "dm":
+      return `https://public-reporting-data.${env}.login.gov`;
+    default:
+      return `https://public-reporting-data.${env}.identitysandbox.gov`;
+  }
+}
+
 interface PathParameters {
   reportName: string;
   date: Date;
-  env?: string;
+  env: string;
 }
 
-function path({ reportName, date, env = "prod" }: PathParameters): string {
+function path({ reportName, date, env }: PathParameters): string {
   const year = yearFormat(date);
   const day = yearMonthDayFormat(date);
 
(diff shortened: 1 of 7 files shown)


src/report.ts
```diff
@@ -3,18 +3,31 @@ import { utcFormat } from "d3-time-format";
...
+function domain(env: string): string {
+  switch (env) {
+    case "local":
+      return "";
+    case "prod":
+    case "staging":
+    case "dm":
+      return `https://public-reporting-data.${env}.login.gov`;
+    default:
+      return `https://public-reporting-data.${env}.identitysandbox.gov`;
```
aduth, line 15: Feels a bit sketchy to put an unfiltered query parameter here. Maybe make `local` behavior the default and safelist the sandbox environments like we're doing with the production environments?

aduth (approved):

src/report.ts
zachmargolis: that means that every time one of us adds our own personal sandbox environment, we have to safelist it?

zachmargolis: this also is only for the sanbox environment, so 1) there shouldn't be any PII and 2) errors are usually more ok?

aduth: At the very least could we sanitize it, e.g. `env.replace(/\W/g, '')` ?

Example abuse I'm worried about: `?env=evil.example.com/a`

zachmargolis: Love it, done in 538b0879cb3ff9d468739e9e32a8f6d5082fb3f9

aduth: Also raises a question of whether we could set up a CSP `connect-src` for the domains we're expecting to fetch from.

zachmargolis: That CSP header would have to come from Federalist, and looks like the only custom headers are Cache-Control right now? 

https://federalist.18f.gov/documentation/custom-headers/

aduth: We can do it with a `<meta>` tag too.

https://developers.google.com/web/fundamentals/security/csp#the_meta_tag

zachmargolis: awesome, added in be38bc42222c0dfddd0fb820ba68802d47f8fd97, I had to remove the param sanitization but CSV did error as expected with `?env=evil.com/a`

PR #17, opened 2021-09-14
Include prebuilt design system styles

aduth:
**Why**: So that the build isn't quite so slow. For additional styling, we can import just the required mixins / functions.


zachmargolis (approved):
LGTM 🚀 verified it works locally, thanks!

PR #18, opened 2021-09-14
Pin chart x-axis domain

zachmargolis:
**Why**: so that the chart does not truncate to available
dates (aka show empty sections)

| before | after |
| --- | --- |
| Screen Shot 2021-09-14 at 1 02 01 PM |  Screen Shot 2021-09-14 at 1 01 52 PM |

### DIFF
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -155,6 +155,9 @@ function DailyAuthsReport(): VNode {
         y: {
           tickFormat: format(".1s"),
         },
+        x: {
+          domain: [start, finish],
+        },
         style: {},
         marks: [
           Plot.ruleY([0]),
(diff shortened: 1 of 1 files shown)


aduth (approved):
Makes sense 👍  

Separately, I notice when navigating through entire weeks where there's no data available, the labels on the X-axis don't update. 

1. Go to https://federalist-14721e0a-0772-4dcd-b62f-ee9d20dcc5fd.app.cloud.gov/preview/18f/identity-reporting/margolis-pin-axis/daily-auths-report/
2. Press "Next Week" twice
3. X-axis labels don't change after the second click

zachmargolis:
> Separately, I notice when navigating through entire weeks where there's no data available, the labels on the X-axis don't update.

Good catch! A caching issue because `start` and `finish` changed but `data` stayed the same (empty array), fixed in 4b2cfad600adc4a60051cb3f5e2fa9ba4140a0e6

PR #19, opened 2021-09-14
Show fewer colors on main page, aggregate by agency

zachmargolis:
Based on some Slack feedback from @nickttng that there are a lot of colors by default

This turns the "unfiltered" report into one that aggregates by agency, so that selecting agency is more of a drilldown, so it changes the table as well

But now that I look at this, I think I want to keep the aggregation in the table and maybe leave out the color changing

| mode | before | after |
| --- | --- | -- |
| no agency selected | Screen Shot 2021-09-14 at 3 17 01 PM |  Screen Shot 2021-09-14 at 3 16 49 PM |
| single agency selected [no change] | Screen Shot 2021-09-14 at 3 17 23 PM | Screen Shot 2021-09-14 at 3 17 35 PM |

### DIFF
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -206,10 +206,7 @@ function DailyAuthsReport(): VNode {
                   thresholds: utcDay,
                 },
                 fill: agency ? "friendly_name" : "steelblue",
-                title: (bin: ProcessedResult[]) => {
-                  const d = bin[0];
-                  return d && (agency ? d.friendly_name : d.agency);
-                },
+                title: agency ? (bin: ProcessedResult[]) => bin[0]?.friendly_name : undefined,
                 filter: (d: ProcessedResult) => d.ial === ial && (!agency || d.agency === agency),
               }
             )
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -68,11 +68,11 @@ function loadData(
 const yearMonthDayFormat = utcFormat("%Y-%m-%d");
 
 function tabulate(
-  results: ProcessedResult[],
+  results?: ProcessedResult[],
   filterAgency?: string,
   filterIal?: number
 ): TableData {
-  const filteredResults = results.filter(
+  const filteredResults = (results || []).filter(
     (d) => (!filterAgency || d.agency === filterAgency) && (!filterIal || d.ial === filterIal)
   );
 
@@ -120,8 +120,8 @@ function tabulate(
   };
 }
 
-function tabulateSumByAgency(results: ProcessedResult[], filterIal?: number): TableData {
-  const filteredResults = results.filter((d) => !filterIal || d.ial === filterIal);
+function tabulateSumByAgency(results?: ProcessedResult[], filterIal?: number): TableData {
+  const filteredResults = (results || []).filter((d) => !filterIal || d.ial === filterIal);
 
   const days = Array.from(new Set(filteredResults.map((d) => d.date.valueOf())))
     .sort((a, b) => a - b)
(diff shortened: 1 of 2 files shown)


src/daily-auths-report.tsx
```diff
@@ -120,6 +120,38 @@ function tabulate(
...
+
+  const header = ["Agency", "IAL", ...days.map(yearMonthDayFormat), "Total"];
+
+  const body = Array.from(rolledup)
+    .sort(([agencyA], [agencyB]) => ascending(agencyA, agencyB))
+    .flatMap(([agency, ials]) =>
+      Array.from(ials).map(([ial, ialDays]) => {
+        const dayCounts = days.map((date) => ialDays.get(yearMonthDayFormat(date)) || 0);
+
+        return [agency, String(ial), ...dayCounts, dayCounts.reduce((d, total) => d + total, 0)];
```
aduth, line 146: Why does only the `ial` number need to be string-ified, and not the subsequent cells? Or does it?

src/daily-auths-report.tsx
```diff
@@ -173,7 +205,7 @@ function DailyAuthsReport(): VNode {
                   value: "date",
                   thresholds: utcDay,
                 },
-                fill: agency ? "friendly_name" : "agency",
+                fill: agency ? "friendly_name" : "steelblue",
                 title: (bin: ProcessedResult[]) => {
```
aduth, line 209: Should we be returning a title anymore for the "All" agencies?

src/daily-auths-report.tsx
```diff
@@ -190,10 +222,13 @@ function DailyAuthsReport(): VNode {
   return (
     <div>
       <div className="chart-wrapper" ref={ref} />
-      <Table data={tabulate(data || [], agency, ial)} numberFormatter={format(",")} />
+      <Table
+        data={agency ? tabulate(data || [], agency, ial) : tabulateSumByAgency(data || [], ial)}
```
aduth, line 226: Maybe `results` should be an optional argument?

```suggestion
        data={agency ? tabulate(data, agency, ial) : tabulateSumByAgency(data, ial)}
```

aduth (approved):

src/daily-auths-report.tsx
zachmargolis: Numbers get right aligned and formatted with commas, and I figured IAL didn't need that treatment

nickttng (approved):
I hadn't anticipated my comment to be taken as a feedback. 😂

The after version of "No agency selected" makes sense to me while keeping the "single agency selected" as is

src/daily-auths-report.tsx
zachmargolis: Probably not! good catch

zachmargolis:
Related, I'd love to make the table linkable, ex clicking on an agency in the table updates the form to select that specific agency, unsure of a good way to do that... will come back to the idea in a future PR

src/daily-auths-report.tsx
zachmargolis: Updated in d72a3828a891f5d514e7ac1036678ccf1a8ab01c

src/daily-auths-report.tsx
zachmargolis: Good idea, updated in 81de3fa60db6c6c20c85a9d5303c4b991c838a55

zachmargolis:
Ok so I have changed my mind... I think it's nice to have a visual breakdown of the agencies on the first page, but still keep to fewer colors... so maybe more like this, gonna give this a shot, it's essentially two different pages now

IMG_0017

zachmargolis:
Added in faceting by agency when all agencies are showing

Screen Shot 2021-09-15 at 10 23 26 AM

PR #20, opened 2021-09-14
Move report outside of the <form> tag

zachmargolis:
**Why**: logically it does not make sense for the
form to contain the plot, and also the form was
limiting the width of the data table

Diff best viewed with [whitespace hidden](https://github.com/18F/identity-reporting/compare/margolis-full-width?w=1)

| before | after |
| --- | --- |
|  Screen Shot 2021-09-14 at 4 56 22 PM | Screen Shot 2021-09-14 at 4 56 15 PM |

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,15 +21,14 @@
     "@typescript-eslint/no-unused-vars": "error",
     "import/no-extraneous-dependencies": ["error", { "devDependencies": ["**/*.test.*", "./*"] }],
     "jsx-a11y/label-has-associated-control": ["error", { "assert": "either" }],
-    "react/jsx-filename-extension": ["error", { "extensions": [".tsx"] }],
-    "react/jsx-fragments": "off"
+    "react/jsx-filename-extension": ["error", { "extensions": [".tsx"] }]
   },
   "settings": {
     "import/resolver": "eslint-import-resolver-typescript",
     "react": {
       "pragma": "h",
       "fragment": "Fragment",
-      "version": "16.0"
+      "version": "16.2"
     }
   },
   "ignorePatterns": ["_site/**"]
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,8 @@
     "@typescript-eslint/no-unused-vars": "error",
     "import/no-extraneous-dependencies": ["error", { "devDependencies": ["**/*.test.*", "./*"] }],
     "jsx-a11y/label-has-associated-control": ["error", { "assert": "either" }],
-    "react/jsx-filename-extension": ["error", { "extensions": [".tsx"] }]
+    "react/jsx-filename-extension": ["error", { "extensions": [".tsx"] }],
+    "react/jsx-fragments": "off"
   },
   "settings": {
     "import/resolver": "eslint-import-resolver-typescript",
--- a/src/report-filter-controls.tsx
+++ b/src/report-filter-controls.tsx
@@ -88,92 +88,94 @@ function ReportFilterControls({
   }
 
   return (
-    <div className="usa-form">
-      <form ref={formRef} onChange={update}>
-        <div>
-          <label>
-            Start
-            <input
-              type="date"
-              name="start"
-              value={yearMonthDayFormat(start)}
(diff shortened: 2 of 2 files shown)


.eslintrc.json
```diff
@@ -21,7 +21,8 @@
     "@typescript-eslint/no-unused-vars": "error",
     "import/no-extraneous-dependencies": ["error", { "devDependencies": ["**/*.test.*", "./*"] }],
     "jsx-a11y/label-has-associated-control": ["error", { "assert": "either" }],
-    "react/jsx-filename-extension": ["error", { "extensions": [".tsx"] }]
+    "react/jsx-filename-extension": ["error", { "extensions": [".tsx"] }],
+    "react/jsx-fragments": "off"
```
aduth, line 25: I see this was logging an error:

```
identity-reporting/src/report-filter-controls.tsx
  91:5  error  Fragments are only supported starting from React v16.2. Please disable the `react/jsx-fragments` rule in ESLint settings or upgrade your version of React  react/jsx-fragments
```

Alternatively, we could change the configured `version` below from `16.0` to `16.2`?

I picked `16.0` based on [official defaults for Preact](https://github.com/preactjs/eslint-config-preact/blob/d9c992501fb9e59d49167298e18fbf3145210abb/index.js#L60-L61), though it seems that `16.2` works perfectly fine, at least for now 🤷

src/report-filter-controls.tsx
```diff
@@ -88,92 +88,94 @@ function ReportFilterControls({
...
-              name="ial"
-              value="1"
-              checked={ial === 1}
-              className="usa-radio__input"
-            />
-            <label htmlFor="ial-1" className="usa-radio__label">
-              IAL 1
+    <>
+      <div className="usa-form">
+        <form ref={formRef} onChange={update}>
```
aduth, line 93: While we're here: Do we need the `div` wrapper, or can we collapse that to the form?

```suggestion
      <form ref={formRef} onChange={update} className="usa-form">
```

aduth (approved):
Another refactoring thought may be to pull up some of the route parameter handling to the "Route" component so we don't need to [shuffle those props along](https://github.com/18F/identity-reporting/blob/89ac506cb22a92b60b55c7ef77566b8f68d8de4e/src/routes/report.tsx#L14-L16), and/or create a context component separate from the form which just holds the values, from which both the form and table are consumers.

e.g.

```tsx
function ReportRoute({
  start: startParam,
  finish: finishParam,
  ial: ialParam,
  agency,
  env: envParam,
}: ReportRouteProps): VNode {
  const start = (startParam ? yearMonthDayParse(startParam) : null) || startOfPreviousWeek;
  const finish = (finishParam ? yearMonthDayParse(finishParam) : null) || endOfPreviousWeek;
  const ial = (parseInt(ialParam || "", 10) || DEFAULT_IAL) as 1 | 2;
  const env = envParam || DEFAULT_ENV;

  return (
    <ReportContextProvider start={start} finish={finish} agency={agency} ial={ial} env={env}>
      <ReportFilterControls />
      <DailyAuthsReport />
    </ReportContextProvider>
  );
}
```

.eslintrc.json
zachmargolis: Oh that makes sense, I thought the error was just wrong since we use preact instead of react 🤦 will update

zachmargolis: 6642360cb79d08ae63a5faf2a91cdd9407149d6d

src/report-filter-controls.tsx
zachmargolis: great catch! 3676bfb6ed3aaf8a93f4d4afd143e4bf0755fd0c

zachmargolis:
> Another refactoring thought may be to pull up some of the route parameter handling to the "Route" component so we don't need to [shuffle those props along](https://github.com/18F/identity-reporting/blob/89ac506cb22a92b60b55c7ef77566b8f68d8de4e/src/routes/report.tsx#L14-L16), and/or create a context component separate from the form which just holds the values, from which both the form and table are consumers.

Big fan of this idea, will follow up in another PR

PR #21, opened 2021-09-15
Reorganized ReportFilterControls and its context

zachmargolis:
Based on this comment: https://github.com/18F/identity-reporting/pull/20#pullrequestreview-755060544


src/report-filter-controls.tsx
```diff
@@ -1,18 +1,18 @@
...
 import { utcWeek, CountableTimeInterval } from "d3-time";
 import { route } from "./router";
 
 const yearMonthDayFormat = utcFormat("%Y-%m-%d");
-const yearMonthDayParse = utcParse("%Y-%m-%d");
 const DEFAULT_IAL = 1;
 const DEFAULT_ENV = "prod";
 
 const endOfPreviousWeek = utcWeek.floor(new Date());
 const startOfPreviousWeek = utcWeek.floor(new Date(endOfPreviousWeek.valueOf() - 1));
```
zachmargolis, line 12: These are now duplicated, could maybe move these and just set the defaults in the context as `new Date()` ? or `null as Date` or something nefarious?

zachmargolis: went ahead and simplified in 0c7f4fca4c07c1e799be097039e18b446497c8f8

src/routes/report.tsx
```diff
@@ -11,11 +20,30 @@ export interface ReportRouteProps {
...
+  const ial = (parseInt(ialParam || "", 10) || DEFAULT_IAL) as 1 | 2;
+  const env = envParam || DEFAULT_ENV;
+  const [allAgencies, setAllAgencies] = useState([] as string[]);
+
   return (
-    <ReportFilterControls {...props}>
+    <ReportFilterControlsContext.Provider
+      value={{ path, start, finish, ial, agency, env, setAllAgencies }}
+    >
+      <ReportFilterControls agencies={allAgencies} />
```
aduth, line 44: The handling of "agencies" feels a bit fragmented, where we store the source of truth in the route, pass the state updater to the context provider, and the current value to the filter controls.

I'd wonder if the context should be handling all of this, and making both `setAllAgencies` and `agencies` available through the context.

e.g.

```tsx
function ReportFilterControlsContextProvider({
  start,
  finish,
  ial,
  setAllAgencies,
  env,
  children,
}: ReportFilterControlsContextProviderProps) {
  const start = (startParam ? yearMonthDayParse(startParam) : null) || startOfPreviousWeek;
  const finish = (finishParam ? yearMonthDayParse(finishParam) : null) || endOfPreviousWeek;
  const ial = (parseInt(ialParam || "", 10) || DEFAULT_IAL) as 1 | 2;
  const env = envParam || DEFAULT_ENV;
  const [allAgencies, setAllAgencies] = useState([] as string[]);

  return (
    <ReportFilterControlsContext.Provider
      value={{ start, finish, ial, env, allAgencies, setAllAgencies }}
    >
      {children}
    </ReportFilterControlsContext.Provider>
  );
}
```

(This might also help for what you mentioned at https://github.com/18F/identity-reporting/pull/21#discussion_r709490268 )

Optionally, we could go even further to have a separate context just for managing "agencies", since it feels a bit like it's not explicitly tied to filtering.

e.g.

```tsx
<AgenciesContextProvider>
  <ReportFilterControlsContextProvider start={start} finish={finish} ial={ial} agency={agency} env={env}>
    <ReportFilterControls />
    <DailyAuthsReport />
  </ReportFilterControlsContextProvider>
</AgenciesContextProvider>
```

src/routes/report.tsx
```diff
@@ -11,11 +20,30 @@ export interface ReportRouteProps {
...
+  const start = (startParam && yearMonthDayParse(startParam)) || startOfPreviousWeek;
+  const finish = (finishParam && yearMonthDayParse(finishParam)) || endOfPreviousWeek;
+  const ial = (parseInt(ialParam || "", 10) || DEFAULT_IAL) as 1 | 2;
+  const env = envParam || DEFAULT_ENV;
+  const [allAgencies, setAllAgencies] = useState([] as string[]);
+
   return (
-    <ReportFilterControls {...props}>
+    <ReportFilterControlsContext.Provider
+      value={{ path, start, finish, ial, agency, env, setAllAgencies }}
```
aduth, line 42: I kinda hoped we wouldn't have to pass around `path` but instead pull it directly from `preact-router`'s internal context. I saw there's an [exported `useRouter`](https://github.com/preactjs/preact-router/blob/24ffb40c8df2a455bc9518aa0ffdb7c5a1878968/src/index.js#L270) which would give us access to this, but it's not yet published 😞

src/report-filter-controls.tsx
```diff
@@ -170,12 +141,10 @@ function ReportFilterControls({
         </div>
         {env !== DEFAULT_ENV && <input type="hidden" name="env" value={env} />}
       </form>
-      <ReportFilterControlsContext.Provider value={filterControls}>
-        {children}
-      </ReportFilterControlsContext.Provider>
+      {children}
```
aduth, line 144: We don't need `children` anymore, do we?

src/routes/report.tsx
zachmargolis: I like the idea of the separate agencies context, will give that a shot

src/routes/report.tsx
zachmargolis: 😩

src/report-filter-controls.tsx
zachmargolis: I guess not! I thought it was convention/habit to always allow children in a component

src/routes/report.tsx
zachmargolis: added AgenciesContext in c69df8e919276d801e7eed83259041aa261420aa

src/routes/report.tsx
zachmargolis: Well I decided that `path` didn't need to live in the filter context so I did move it out in fe3169f9b30a1f3d0b3ba8ad9073ca7406b48f35

aduth: > Well I decided that `path` didn't need to live in the filter context so I did move it out in [fe3169f](https://github.com/18F/identity-reporting/commit/fe3169f9b30a1f3d0b3ba8ad9073ca7406b48f35)

That does seem better 👍

src/agencies-context.ts
```diff
@@ -0,0 +1,14 @@
+import { createContext } from "preact";
```
aduth, line 1: Since we'll possibly have more context, maybe an opportunity for some folder organization? As `src/context/agencies-context.ts` for example.

src/agencies-context.ts
```diff
@@ -0,0 +1,14 @@
...
+
+interface AgenciesContextValues {
+  agencies: string[];
+  setAgencies: StateUpdater<string[]>;
+}
+
+const AgenciesContext = createContext({
+  agencies: [],
+  setAgencies: () => null,
+} as AgenciesContextValues);
```
aduth, line 12: That we have to pass in `agencies` and `setAgencies` from the rendering component feels a bit odd to me, as opposed to where the context would manage this itself to be self-contained, e.g. with a custom provider component:

```tsx
function AgenciesProvider({ children }) {
  const [agencies, setAgencies] = useState([]);
  return <AgenciesContext.Provider value={{ agencies, setAgencies }}>{children}</AgenciesContext.Provider>;
}
```

That being said, I've found competing advice on this sort of pattern, so not something I feel too strongly about right now.

aduth: It's also something I was curious to explore for the ReportFilterControlsContext:

https://github.com/18F/identity-reporting/compare/margolis-shuffle-props...aduth-report-filter-controls-context?expand=1

aduth (approved):

src/agencies-context.ts
zachmargolis: Updated in 32c68c59d5c8c526619032ca6cfb9be528d719fc for just the agencies provider

PR #22, opened 2021-09-15
Update Nav

zachmargolis:
we only have one page...so some of this may be overkill 😬 

| before | after |
| --- | --- |
| Screen Shot 2021-09-15 at 12 32 41 PM | Screen Shot 2021-09-15 at 12 32 26 PM |

And mobile too

| closed | expanded |
| --- | -- |
| Screen Shot 2021-09-15 at 12 38 24 PM | Screen Shot 2021-09-15 at 12 38 33 PM |

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -12,14 +12,14 @@ render(
     <header className="usa-header usa-header--extended">
       <div className="usa-navbar">
         <div className="usa-logo">
-          <a href="/" title="Home" aria-label="Home">
+          <Link href="/" title="Home" aria-label="Home">
             <img
               src="/node_modules/identity-style-guide/dist/assets/img/login-gov-logo.svg"
               className="usa-logo__img"
               alt="Login.gov"
             />
             <em className="usa-logo__text">Data</em>
-          </a>
+          </Link>
         </div>
         <button className="usa-menu-btn" type="button">
           Menu
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -1,11 +1,11 @@
 import "./css/style.scss";
 
 import { render } from "preact";
-import { banner, accordion } from "identity-style-guide";
+import { banner, accordion, navigation } from "identity-style-guide";
 import { Link } from "./router";
 import { Routes } from "./routes";
 
-[banner, accordion].forEach((component) => component.on());
+[banner, accordion, navigation].forEach((component) => component.on());
 
 render(
   <div>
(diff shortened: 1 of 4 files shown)


src/main.tsx
```diff
@@ -1,22 +1,50 @@
...
-      </div>
-    </nav>
+      <nav className="usa-nav">
+        <div className="usa-nav__inner">
+          <button className="usa-nav__close" type="button">
+            <img src="/node_modules/identity-style-guide/dist/assets/img/close.svg" alt="Close" />
+          </button>
+          <ul className="usa-nav__primary usa-accordion">
+            <li className="usa-nav__primary-item">
+              <Link href="/" activeClassName="usa-current">
```
zachmargolis, line 35: `activeClassName` [as documented in preact-router](https://github.com/preactjs/preact-router) doesn't seem to work? It just gets splatted into the dom as `activeclassname="usa-current"`... but the react types say that it should exist?

https://github.com/preactjs/preact-router/blob/24ffb40c8df2a455bc9518aa0ffdb7c5a1878968/index.d.ts#L75-L77

🤷

src/main.tsx
```diff
@@ -1,22 +1,50 @@
...
 
 render(
   <div>
-    <nav>
-      <div>
-        <Link href="/">Home</Link>
+    <header className="usa-header usa-header--extended">
+      <div className="usa-navbar">
+        <div className="usa-logo">
+          <a href="/" title="Home" aria-label="Home">
```
aduth, line 15: `<Link>` here?

src/main.tsx
aduth: Looks familiar: https://github.com/preactjs/preact-router/issues/379

aduth: Maybe we can recreate it like `<Link className={path === '/' ? 'usa-current' : null }>`

src/main.tsx
```diff
@@ -1,22 +1,50 @@
...
 
-[banner, accordion].forEach((component) => component.on());
+[banner, accordion, navigation].forEach((component) => component.on());
 
 render(
   <div>
-    <nav>
-      <div>
-        <Link href="/">Home</Link>
+    <header className="usa-header usa-header--extended">
```
aduth, line 12: Worth moving into a component?

aduth (commented):
We could publish new version of `identity-style-guide` if you want https://github.com/18F/identity-style-guide/pull/249 ?

zachmargolis:
> We could publish new version of `identity-style-guide` if you want [18F/identity-style-guide#249](https://github.com/18F/identity-style-guide/pull/249) ?

No need to rush a release just for this? It'll get here eventually

aduth:
> > We could publish new version of `identity-style-guide` if you want [18F/identity-style-guide#249](https://github.com/18F/identity-style-guide/pull/249) ?
> 
> No need to rush a release just for this? It'll get here eventually

I don't mind either way. If you're not worried about it, it can wait 👍

src/main.tsx
zachmargolis: done in 40d5a9cfd38d006b2ec14ff57736c7d5688f689a

src/main.tsx
zachmargolis: Since this is top level `render` we don't have access to path? But we wouldn't be able to get that directly from a component either?

aduth: Hm, that's frustrating 😕  

I'd wonder if we could create our own context to hold the path, using the `Router` `onChange` prop:

https://github.com/preactjs/preact-router/blob/61671b38fd5e371e886a6b33db2c66df7a6444db/src/index.js#L235

At a glance, I'm not sure the best way to meaningfully tie that to a specific "path" or route component.

We could also defer this for now and handle it later.

Aside: I also realize looking at the internal implementation: The way we're [using `path` to build a new path fragment](https://github.com/18F/identity-reporting/blob/602bb519caca86d484eecdf28a31d336eb757a17/src/report-filter-controls.tsx#L77) may not always work if `path` is actually the path pattern and could be something like [`/d/:optional?/:params?`](https://github.com/preactjs/preact-router#handling-urls).

aduth (approved):

src/main.tsx
zachmargolis: Yeah we may need to come back to this --- Also I noticed routing to `/` does not change the path... but it does change the display, unsure how to fix that right now

src/main.tsx
zachmargolis: 🤷 it's only used once and doesn't contain any logic? Happy to move if you feel strongly, will leave in place for now

src/main.tsx
aduth: > Also I noticed routing to `/` does not change the path... but it does change the display, unsure how to fix that right now

If I follow correctly, I think this is fixed in #26 by introducing the route handler for a "Home" route.

PR #23, opened 2021-09-15
Add some details about the report, and a title

zachmargolis:
Adds some new details

| closed | expanded |
| --- | --- |
| Screen Shot 2021-09-15 at 1 43 20 PM | Screen Shot 2021-09-15 at 1 43 26 PM |

Draws inspiration from the [VA Performance Dashboard](https://www.va.gov/performance-dashboard/)

Screen Shot 2021-09-15 at 1 45 20 PM

### DIFF
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -259,9 +259,13 @@ function DailyAuthsReport(): VNode {
   return (
     <div>
       <h2>Daily Auths Report</h2>
-      <details>
-        <summary>How is this measured?</summary>
-        <div className="usa-prose">
+      <div className="usa-accordion usa-accordion--bordered margin-bottom-2">
+        <h3 className="usa-accordion__heading">
+          <button className="usa-accordion__button" aria-controls="how-is-it-measured" aria-expanded="false">
+            How is this measured?
+          </button>
+        </h3>
+        <div className="usa-prose usa-accordion__content" id="how-is-it-measured" hidden>
           <p>
             <strong>Timing: </strong>
             All data is collected, grouped, and displayed in the UTC timezone.
@@ -272,7 +276,7 @@ function DailyAuthsReport(): VNode {
             twice will count twice. It does not de-duplicate users or provide unique auths.
           </p>
         </div>
-      </details>
+      </div>
       <PlotComponent
         plotter={() => plot({ data, ial, agency, start, finish })}
         inputs={[data, ial, agency, start.valueOf(), finish.valueOf()]}
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -261,7 +261,12 @@ function DailyAuthsReport(): VNode {
       <h2>Daily Auths Report</h2>
       <div className="usa-accordion usa-accordion--bordered margin-bottom-2">
         <h3 className="usa-accordion__heading">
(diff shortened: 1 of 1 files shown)


aduth (commented):
Consider an [Accordion](https://design.login.gov/components/accordions/) ?

zachmargolis:
> Consider an [Accordion](https://design.login.gov/components/accordions/) ?

But it needs more custom JS! I love that there's built-in behavior from details/summary

Will give this a shot

zachmargolis:
> > Consider an [Accordion](https://design.login.gov/components/accordions/) ?
> 
> But it needs more custom JS! I love that there's built-in behavior from details/summary
> 
> Will give this a shot

used the component in 0c58279b7d422a936d22d49857fdf31a5e275b05, screenshot:

Screen Shot 2021-09-15 at 2 04 28 PM

src/daily-auths-report.tsx
```diff
@@ -258,6 +258,30 @@ function DailyAuthsReport(): VNode {
 
   return (
     <div>
+      <h2>Daily Auths Report</h2>
+      <div className="usa-accordion usa-accordion--bordered margin-bottom-2">
```
aduth, line 262: Something we'll want to keep an eye out for: Some USWDS components have "initialization" logic, which only runs when `component.on()` is called. For the accordion, all the initialization does is [handle if `aria-expanded="true"`](https://github.com/uswds/uswds/blob/e1ed21783b089a2831504971b3426fbea1b5b752/src/js/components/accordion.js#L86-L91). We're not using that (currently), so it doesn't affect us. Since the primary toggling behavior is handled by event delegation, it works fine as-is.

If we'd need to run initialization behavior, it'd be something where I expect we'd have to do something like this instead:

```js
function Accordion() {
  const accordionRef = useRef();
  useEffect(() => {
    accordion.on(accordionRef.current?);
  }, []);

  return <div ref={accordionRef} className="usa-accordion" />;
}
```

This varies by component as well, since other components rely heavily on initialization.

(Another thing I'm [keen to improve](https://github.com/uswds/uswds/issues/2884#issuecomment-872238628))

src/daily-auths-report.tsx
```diff
@@ -258,6 +258,30 @@ function DailyAuthsReport(): VNode {
 
   return (
     <div>
+      <h2>Daily Auths Report</h2>
```
aduth, line 261: I'd consider moving this up the page and using `<h1>`. Since there's currently no h1, it causes a few accessibility issues:

- https://dequeuniversity.com/rules/axe/3.5/page-has-heading-one
- https://dequeuniversity.com/rules/axe/4.2/heading-order

aduth (approved):

src/daily-auths-report.tsx
zachmargolis:
```suggestion
      <h1>Daily Auths Report</h1>
```

zachmargolis: Just saw this comment in the image fixing PR, will just remove this here: https://github.com/18F/identity-reporting/pull/26/files#r710159268

src/daily-auths-report.tsx
```diff
@@ -258,6 +258,30 @@ function DailyAuthsReport(): VNode {
 
   return (
     <div>
+      <h1>Daily Auths Report</h1>
```
zachmargolis, line 261:
```suggestion
```

PR #24, opened 2021-09-16
Add responsive resizing of charts

zachmargolis:
Super basic resizing of the table. It relies on preact to collect duplicate events, so maybe there is some additional debouncing we could add to simplify extra renders if we wanted... (or if we did the SVG chart rendering more directly, we could manage the elements do they wouldn't have to be torn down each time... it's fast enough for now and not causing any problems)

~**Next up**: making the table scrollable, it gets cut off when the breakpoints get too small~ (done, 3918f5a15dd77b4dd5b756aece92e1ed6e9c9a49)

| before | after |
| --- | --- |
| Screen Shot 2021-09-15 at 5 09 28 PM | Screen Shot 2021-09-15 at 5 09 44 PM |
| | Screen Shot 2021-09-15 at 5 10 01 PM |
| | Screen Shot 2021-09-15 at 5 10 09 PM |

### DIFF
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -1,5 +1,5 @@
 import { VNode, ComponentChildren } from "preact";
-import { useContext, useEffect, useRef, Inputs } from "preact/hooks";
+import { useContext, useEffect, useRef, Inputs, useState } from "preact/hooks";
 import { utcDays, utcDay } from "d3-time";
 import * as Plot from "@observablehq/plot";
 import { format } from "d3-format";
@@ -160,6 +160,7 @@ function plot({
   agency,
   ial,
   facetAgency,
+  width,
 }: {
   start: Date;
   finish: Date;
@@ -167,9 +168,11 @@ function plot({
   agency?: string;
   ial: number;
   facetAgency?: boolean;
+  width?: number;
 }): HTMLElement {
   return Plot.plot({
     height: facetAgency ? new Set((data || []).map((d) => d.agency)).size * 60 : undefined,
+    width,
     y: {
       tickFormat: format(".1s"),
     },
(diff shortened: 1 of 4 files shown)


aduth:
Since it's just rendering an `<svg>`, could we tell the plotting library to use percentage-based width instead? `resize` events are pretty performance-intensive.

Looking at the [docs](https://www.npmjs.com/package/@observablehq/plot), maybe:

```js
Plot.plot({
  // ...
  style: {
    width: '80%',
    height: 'auto',
    // max-width: '...'
  },
});
```

aduth (approved):

src/daily-auths-report.tsx
```diff
@@ -256,16 +262,25 @@ function DailyAuthsReport(): VNode {
...
+    listener();
+
+    window.addEventListener("resize", listener);
+    return () => window.removeEventListener("resize", listener);
+  });
+
   return (
-    <div>
+    <div ref={ref}>
       <PlotComponent
```
aduth, line 276: I hadn't noticed 'til now, but it'd be nice to split this file up. Typically I'd only expect one component per file, so pulling out `Plot` to its own file might be a nice way to handle that?

zachmargolis:
> Since it's just rendering an `<svg>`, could we tell the plotting library to use percentage-based width instead? `resize` events are pretty performance-intensive.

Unfortunately, d3 underneath converts everything to pixel space... using `width: 100%` gets gnarly errors like this:

```
attr.js:17 Error: <svg> attribute viewBox: Expected number, "0 0 100% 1800".
```

src/daily-auths-report.tsx
zachmargolis: I knew you'd comment on that! 😝 My thinking was more along the lines of "if we aren't going to reference it from a second file, it's fine to keep here for now", but I will move PlotComponent out

zachmargolis: 88e2fa3a29aba7fb6011aedcb68a8f33ae17f3d3

aduth:
> Unfortunately, d3 underneath converts everything to pixel space... using `width: 100%` gets gnarly errors like this:

I was thinking using the `style` option, not the `width` option.

zachmargolis:
> I was thinking using the `style` (`style.width`) option, not the `width` option.

Just gave this a shot...does not seem to resize

```diff
 }): HTMLElement {
   return Plot.plot({
     height: facetAgency ? new Set((data || []).map((d) => d.agency)).size * 60 : undefined,
-     width,
+    style: {
+      width: '100%',
+    },
```

PR #25, opened 2021-09-16
Update tick formats on chart

zachmargolis:
- Add an extra significant figure
- Drop ".0" because "0.0" is redundant

For the most part d3-format's [SI prefix](https://github.com/d3/d3-format#api-reference) works, but converts "0" to "0.0" which looks really weird on graphs, especially when we only have whole-number auths in this case.

| notes | screenshot |
| --- | --- |
| before, 1 sig fig, has redundant breaks | Screen Shot 2021-09-15 at 6 53 31 PM |
| in progress, 2 sig figs, has "0.0" | Screen Shot 2021-09-15 at 6 53 40 PM |
| final product, 2 sig figs except drops trailing zeroes |  Screen Shot 2021-09-15 at 6 53 52 PM |

### DIFF
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -153,6 +153,9 @@ function tabulateSumByAgency(results?: ProcessedResult[], filterIal?: number): T
   };
 }
 
+const formatSIPrefix = format(".2s");
+const formatSIDropTrailingZeroes = (d: number): string => formatSIPrefix(d).replace(/\.0+/, "");
+
 function plot({
   start,
   finish,
@@ -171,7 +174,7 @@ function plot({
   return Plot.plot({
     height: facetAgency ? new Set((data || []).map((d) => d.agency)).size * 60 : undefined,
     y: {
-      tickFormat: format(".1s"),
+      tickFormat: formatSIDropTrailingZeroes,
     },
     x: {
       domain: [start, finish],
@@ -277,4 +280,4 @@ function DailyAuthsReport(): VNode {
 }
 
 export default DailyAuthsReport;
-export { tabulate, tabulateSumByAgency, loadData };
+export { tabulate, tabulateSumByAgency, loadData, formatSIDropTrailingZeroes };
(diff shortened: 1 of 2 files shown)


src/daily-auths-report.tsx
```diff
@@ -153,6 +153,9 @@ function tabulateSumByAgency(results?: ProcessedResult[], filterIal?: number): T
   };
 }
 
+const formatSIPrefix = format(".2s");
+const formatSIDropTrailingZeroes = (d: number): string => formatSIPrefix(d).replace(/\.0+/, "");
```
aduth, line 157: Is it intended to drop the decimal numbers if there's a non-zero after the zero, e.g. `1000.09` ? 

If not:

```suggestion
const formatSIDropTrailingZeroes = (d: number): string => formatSIPrefix(d).replace(/\.0+$/, "");
```

aduth (approved):

src/daily-auths-report.tsx
aduth: Or are there any digits ever after the `.0` ? 🤔  If yes, then this could cause some strangeness, e.g.

```
> '1000.09'.replace(/\.0+/, '')
'10009'
```

I'd expect one of:

```suggestion
const formatSIDropTrailingZeroes = (d: number): string => formatSIPrefix(d).replace(/\.0/, "");
```

```suggestion
const formatSIDropTrailingZeroes = (d: number): string => formatSIPrefix(d).replace(/\.0.+$/, "");
```

```suggestion
const formatSIDropTrailingZeroes = (d: number): string => formatSIPrefix(d).replace(/\.0+$/, "");
```

zachmargolis: The `.2` to d3-format says 2 significant digits, so `1000.09` will convert to `1.0k` always.

For safety, I'll add the anchored regex

zachmargolis: The anchor got messed up by the SI suffix (the "k" character in tests) so I'm going to remove this... precision + tests should cover it for now

PR #26, opened 2021-09-16
Fix images rendered by application

aduth:
**Why**: Vite requires static assets to be imported. It works locally only because node_modules exists in the working directly. But this folder is not created in the built environments.

### DIFF
--- a/bin/generate-routes.js
+++ b/bin/generate-routes.js
@@ -1,9 +1,8 @@
 import { mkdir, copyFile } from "fs/promises";
 import { join } from "path";
-import { ROUTES } from "../src/routes";
 
 Promise.all(
-  Object.keys(ROUTES).map(async (path) => {
+  ["/daily-auths-report/"].map(async (path) => {
     const dir = join("_site", path);
     await mkdir(dir, { recursive: true });
     await copyFile("_site/index.html", join(dir, "index.html"));
--- a/src/header.tsx
+++ b/src/header.tsx
@@ -1,0 +1,50 @@
+import { VNode } from "preact";
+import logoURL from "../node_modules/identity-style-guide/dist/assets/img/login-gov-logo.svg";
+import closeURL from "../node_modules/identity-style-guide/dist/assets/img/close.svg";
+import { Link } from "./router";
+
+interface HeaderProps {
+  path: string;
+}
+
+function Header({ path }: HeaderProps): VNode {
+  return (
+    <header className="usa-header usa-header--extended">
+      <div className="usa-navbar">
+        <div className="usa-logo">
+          <Link href="/" title="Home" aria-label="Home">
+            <img src={logoURL} className="usa-logo__img" alt="Login.gov" />
+            <em className="usa-logo__text">Data</em>
+          </Link>
+        </div>
+        <button className="usa-menu-btn" type="button">
+          Menu
+        </button>
+      </div>
+      <nav className="usa-nav">
+        <div className="usa-nav__inner">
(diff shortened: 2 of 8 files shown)


bin/generate-routes.js
```diff
@@ -1,9 +1,8 @@
 import { mkdir, copyFile } from "fs/promises";
 import { join } from "path";
-import { ROUTES } from "../src/routes";
 
 Promise.all(
-  Object.keys(ROUTES).map(async (path) => {
+  ["/daily-auths-report/"].map(async (path) => {
```
aduth, line 5: Not super happy about this, but sometimes ESBuild tries (and fails) to parse the SVG, and in other cases Node doesn't know how to handle the static asset imports.

```
> node -r esbuild-register bin/generate-routes.js

identity-reporting/node_modules/identity-style-guide/dist/assets/img/login-gov-logo.svg:1
<?xml version="1.0" encoding="UTF-8"?>
^

SyntaxError: Unexpected token '<'
```

```
> node -r esbuild-register bin/generate-routes.js

identity-reporting/node_modules/esbuild-register/dist/node.js:2699
        throw error;
        ^

Error: Cannot find module '../node_modules/identity-style-guide/dist/assets/img/login-gov-logo.svg?url'
```

There's probably a way to fix this.

src/header.tsx
```diff
@@ -0,0 +1,50 @@
...
+        </button>
+      </div>
+      <nav className="usa-nav">
+        <div className="usa-nav__inner">
+          <button className="usa-nav__close" type="button">
+            <img src={closeURL} alt="Close" />
+          </button>
+          <ul className="usa-nav__primary usa-accordion">
+            <li className="usa-nav__primary-item">
+              <Link href="/" className={path === "/" ? "usa-current" : undefined}>
```
aduth, line 31: Gets the current path into the header by rendering it _within_ the route, not as an ancestor of it.

src/page.tsx
```diff
@@ -0,0 +1,20 @@
...
+  path: string;
+  children?: ComponentChildren;
+  title: string;
+}
+
+function Page({ path, children, title }: PageProps): VNode {
+  return (
+    <>
+      <Header path={path} />
+      <h1>{title}</h1>
```
aduth, line 14: This addresses the comment here: https://github.com/18F/identity-reporting/pull/23#discussion_r710091303

Even better than rendering it per-route, imo.

src/header.tsx
aduth: Hm, this works locally, but not in Federalist.

zachmargolis (approved):
LGTM

src/daily-auths-report.tsx
```diff
@@ -249,7 +249,7 @@ function DailyAuthsReport(): VNode {
 
   return (
     <div ref={ref}>
-      <div className="usa-accordion usa-accordion--bordered margin-bottom-2">
+      <div className="usa-accordion usa-accordion--bordered margin-top-2 margin-bottom-2">
```
zachmargolis, line 252: I snuck in the `margin-top-2` while I was here

src/page.tsx
```diff
@@ -0,0 +1,26 @@
...
+  path: string;
+  children?: ComponentChildren;
+  title: string;
+}
+
+function Page({ path, children, title }: PageProps): VNode {
+  return (
+    <>
+      <Header path={path} />
+      <div className="grid-container">
```
zachmargolis, line 14: I *think* we want the grid container to not include the header? so I stuck it here

PR #27, opened 2021-09-16
Update title formatting

zachmargolis:
- Show exact number for bar
- Show date
- Optionally show app

| notes | screenshot |
| --- | --- |
| across all agencies: exact number + date  |  Screen Shot 2021-09-15 at 7 40 17 PM |
| agency-specific: app name + exact number + date  | Screen Shot 2021-09-15 at 7 40 02 PM |

### DIFF
--- a/src/daily-auths-report.tsx
+++ b/src/daily-auths-report.tsx
@@ -213,10 +213,7 @@ function plot({
               const total = formatWithCommas((bin || []).reduce((sum, d) => sum + d.count, 0));
               const friendlyName = bin[0]?.friendly_name;
 
-              return [
-                agency && `${friendlyName}:`,
-                total, `(${date})`
-              ].filter(Boolean).join(" ");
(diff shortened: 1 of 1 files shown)


src/daily-auths-report.tsx
```diff
@@ -207,7 +208,13 @@ function plot({
               thresholds: utcDay,
             },
             fill: agency ? "friendly_name" : "steelblue",
-            title: agency ? (bin: ProcessedResult[]) => bin[0]?.friendly_name : undefined,
+            title: (bin: ProcessedResult[]) => {
+              const date = yearMonthDayFormat(bin[0].date);
+              const total = formatWithCommas((bin || []).reduce((sum, d) => sum + d.count, 0));
```
aduth, line 213: `bin` would never be falsey would it? Per argument type and the fact we access an index of it on the previous line.

```suggestion
              const total = formatWithCommas(bin.reduce((sum, d) => sum + d.count, 0));
```

aduth (approved):

PR #28, opened 2021-09-16
Add links from table to individual agencies

zachmargolis:
Previously the links were just static text but I think this makes things slightly easier to navigate.

The building of the URL is kind of a kludge.... but I couldn't find a clean way to update the form values other than this 😬 

Screen Shot 2021-09-16 at 2 11 37 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "serve": "vite preview",
     "test:lint": "eslint . --ext .js,.ts,.tsx",
     "test:typecheck": "tsc",
-    "test:unit": "mocha -r esbuild-register -r ./mocha-setup.ts 'src/**/*.test.{js,ts,tsx}'",
+    "test:unit": "mocha -r esbuild-register -r global-jsdom/register -r ./mocha-setup.ts 'src/**/*.test.{js,ts,tsx}'",
     "test": "run-p test:*"
   },
   "dependencies": {
@@ -31,12 +31,15 @@
   },
   "devDependencies": {
     "@18f/eslint-plugin-identity": "^1.0.0",
+    "@testing-library/preact": "^2.0.1",
+    "@testing-library/preact-hooks": "^1.1.0",
     "@types/chai": "^4.2.21",
     "@types/d3-array": "^3.0.1",
     "@types/d3-format": "^3.0.1",
     "@types/d3-time": "^3.0.0",
     "@types/d3-time-format": "^4.0.0",
     "@types/fetch-mock": "^7.3.5",
+    "@types/jsdom": "^16.2.13",
     "@types/mocha": "^9.0.0",
     "@typescript-eslint/eslint-plugin": "^4.29.3",
     "@typescript-eslint/parser": "^4.29.3",
@@ -45,6 +48,8 @@
     "eslint": "^7.32.0",
     "eslint-import-resolver-typescript": "^2.4.0",
     "fetch-mock": "^9.11.0",
+    "global-jsdom": "^8.2.0",
+    "jsdom": "^17.0.0",
     "mocha": "^9.1.0",
     "node-fetch": "^2.6.1",
     "npm-run-all": "^4.1.5",
--- a/src/context/report-filter-context.tsx
+++ b/src/context/report-filter-context.tsx
@@ -10,6 +10,7 @@ interface ReportFilterContextValues {
   ial: 1 | 2;
   agency?: string;
   env: string;
+  setParameters: (params: Record<string, string>) => void;
 }
 
 /**
@@ -34,12 +35,14 @@ const ReportFilterContext = createContext({
   ial: DEFAULT_IAL,
   env: DEFAULT_ENV,
   setParameters: defaultSetParameters,
-} as ReportFilterContextValues & { setParameters: (params: Record<string, string>) => void });
+} as ReportFilterContextValues);
+
+type ReportFilterContextProviderProps = Omit<ReportFilterContextValues, "setParameters">;
 
 function ReportFilterContextProvider({
   children,
   ...rest
-}: { children: ComponentChildren } & ReportFilterContextValues): VNode {
+}: { children: ComponentChildren } & ReportFilterContextProviderProps): VNode {
   return (
     <ReportFilterContext.Provider value={{ ...rest, setParameters: defaultSetParameters }}>
       {children}
--- a/src/context/report-filter-context.tsx
+++ b/src/context/report-filter-context.tsx
@@ -10,7 +10,6 @@ interface ReportFilterContextValues {
   ial: 1 | 2;
   agency?: string;
   env: string;
-  setParameters: (params: Record<string, string>) => void;
 }
 
 /**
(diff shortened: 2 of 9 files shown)


src/url-params.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * preact-router does not parse "+" as spaces the way JS encodes them
+ */
+function pathWithParams(path: string, searchParams: URLSearchParams): string {
+  return `${path}?${searchParams.toString().replace(/\+/g, "%20")}`;
+}
```
zachmargolis, line 6: this is kind of related to our router? but not quite? so I could see an argument for putting it in there?

aduth: I could see it going either way. Maybe even an enhancement to our `Link` wrapper like `<Link path={...} params={...} />`. Or we could go full-"one package per method" and it could be a standalone package via [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) 😄 !

aduth: Or maybe, with #29, this could be a responsibility of `ReportFilterContextProvider` ? Passing a method like `setParameters` to context, which creates the URL to route when set. Then we render a `<button class="usa-button usa-button--unstyled">` instead of a `<Link>`.

aduth (approved):

src/url-params.ts
zachmargolis: good idea, added in 67c45ab6f761d3603fe95cfaeb370d11c65466bb

src/report-filter-controls.tsx
```diff
@@ -40,9 +25,11 @@ function ReportFilterControls({ path }: ReportFilterControlsProps): VNode {
       return;
     }
     const formData = Array.from(new FormData(form)) as string[][];
-    const seachParams = new URLSearchParams(formData);
-    Object.entries(overrideFormData).forEach(([key, value]) => seachParams.set(key, String(value)));
-    route(`${path}?${seachParams.toString().replace(/\+/g, "%20")}`);
+    const searchParams = new URLSearchParams(formData);
```
aduth, line 28: We could use `setParameters` here as well? My thinking with it was also that we could avoid the need for a separate `src/url-params.ts` altogether, since the `pathWithParams` function could either be inlined into `defaultSetParameters` or at least in the `src/context/report-filter-context.tsx` file.

zachmargolis: Gave that a shot in 48d97c2a72c2390a52749f80a60c417ff99db13c and f778d06ef73cd23d836601265e1f1e41f0968879

package.json
```diff
@@ -45,6 +48,8 @@
     "eslint": "^7.32.0",
     "eslint-import-resolver-typescript": "^2.4.0",
     "fetch-mock": "^9.11.0",
+    "global-jsdom": "^8.2.0",
```
aduth, line 51: I'm more familiar with [`jsdom-global`](https://www.npmjs.com/package/jsdom-global) which is a bit more popular as well (based on weekly downloads), but it appears they do the same thing anyways 🤷

src/daily-auths-report.tsx
```diff
@@ -122,7 +122,11 @@ function tabulate(
   };
 }
 
-function tabulateSumByAgency(results?: ProcessedResult[], filterIal?: number): TableData {
+function tabulateSumByAgency(
+  results?: ProcessedResult[],
+  filterIal?: number,
+  setParameters?: (params: Record<string, string>) => void
```
aduth, line 128: No need for this to be optional, is there? Saves us a optional chaining operator below as well.

```suggestion
  setParameters: (params: Record<string, string>) => void
```

src/report-filter-controls.tsx
```diff
@@ -1,48 +1,30 @@
...
+
     const formData = Array.from(new FormData(form)) as string[][];
-    const seachParams = new URLSearchParams(formData);
-    Object.entries(overrideFormData).forEach(([key, value]) => seachParams.set(key, String(value)));
-    route(`${path}?${seachParams.toString().replace(/\+/g, "%20")}`);
+    const searchParams = new URLSearchParams(formData);
+    Object.entries(overrideFormData).forEach(([key, value]) =>
+      searchParams.set(key, String(value))
+    );
+    setParameters(Object.fromEntries(searchParams));
```
aduth, line 27: At a glance, it looks like it can work to just call `Object.fromEntries` on form data directly?

```suggestion
    setParameters(Object.fromEntries(new FormData(form)));
```

src/routes/report-route.tsx
```diff
@@ -40,10 +41,17 @@ function ReportRoute({
...
       <AgenciesContextProvider>
-        <ReportFilterControlsContext.Provider value={{ start, finish, ial, agency, env }}>
-          <ReportFilterControls path={path} />
+        <ReportFilterContextProvider
+          start={start}
+          finish={finish}
+          ial={ial}
+          agency={agency}
+          env={env}
+          setParameters={defaultSetParameters}
```
aduth, line 50: Would be nice if we could just omit the prop to let it take the default.

aduth (approved):

package.json
zachmargolis: I copied these docs:
https://testing-library.com/docs/react-testing-library/setup/#using-without-jest

src/daily-auths-report.tsx
zachmargolis: we can't have required args after optional args so it would change the order of the function a bit

package.json
aduth: Ah, In that case, I see no reason to change 👍

src/routes/report-route.tsx
zachmargolis: Yeah I'll have to make the component take separate attributes than the context

src/daily-auths-report.tsx
aduth: Ah, I guess I assumed this was an object argument.

src/report-filter-controls.tsx
zachmargolis: Gave a shot in 9321f8bbb56ea5aeabe858da7b0ba0b4d8b14e4d

src/routes/report-route.tsx
zachmargolis: updated in 9a5e4aee702354f3b2d12f698c045d0525e0f3f1

src/daily-auths-report.tsx
zachmargolis: object argument makes more sense... updated in 317631463d30761b24ca9d10b2235f7ac91d5662

src/context/report-filter-context.tsx
```diff
@@ -0,0 +1,51 @@
...
+  finish: new Date(),
+  ial: DEFAULT_IAL,
+  env: DEFAULT_ENV,
+  setParameters: defaultSetParameters,
+} as ReportFilterContextValues & { setParameters: (params: Record<string, string>) => void });
+
+function ReportFilterContextProvider({
+  children,
+  ...rest
+}: { children: ComponentChildren } & ReportFilterContextValues): VNode {
```
aduth, line 42: Nit: Passing `setParameters` as a prop is allowed, but it doesn't do anything. If we don't want to make it configurable, we could use the [`Omit` utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys)?

```suggestion
}: { children: ComponentChildren } & Omit<ReportFilterContextValues, 'setParameters'>): VNode {
```

(Which could be extracted e.g. `type ReportFilterContextProviderProps = Omit<ReportFilterContextValues, 'setParameters'>`)

Else I'd expect something below like `{ ...rest, setParameters: setParameters ?? defaultSetParameters }`.

aduth (approved):

src/context/report-filter-context.tsx
zachmargolis: Omit was what I was looking for: f6afa30c9fa9c3f8045dd8acc6637cd7a98e4ae9

PR #31, opened 2021-09-23
Dropoffs Report

zachmargolis:
This is a draft because:
- [x] need to add tests
- [x] need to make a branch for the IDP code that generates this new CSV report (https://github.com/18F/identity-idp/pull/5440)

Screen Shot 2021-09-23 at 8 32 41 AM

Screen Shot 2021-09-23 at 9 40 38 AM


src/daily-dropoffs-report.tsx
```diff
@@ -0,0 +1,376 @@
...
+            {toStepCounts(row).map(({ step, count }, idx, stepCounts) => {
+              let comparedToFirst = 0;
+              if (idx > 0) {
+                comparedToFirst = count / stepCounts[0].count;
+              }
+
+              return (
+                <>
+                  <circle cx={x(step)} cy={y(count)} r="3" fill={color(row.issuer)} />
+                  <text x={x(step)} y={y(count)} font-size="12" dx="3" dy="-3">
```
zachmargolis, line 313: Running `npm run prettier:fix` changed this to be `fontSize` which is wrong :[

src/daily-dropoffs-report.tsx
```diff
@@ -0,0 +1,376 @@
...
+//    * Starts funnel at the welcome screen
+//    */
+//   OVERALL = "overall",
+//   /**
+//    * Starts funnel at the image submission screen
+//    */
+//   BLANKET = "blanket",
+// }
+
+enum Step {
```
zachmargolis, line 29: I can't figure out what the source of this eslint error is... this is the first place that the enum is declared...

```
   29:6   error  'Step' is already declared in the upper scope on line 29 column 6  no-shadow
```

src/daily-dropoffs-report.tsx
zachmargolis:
```
  313:50  error  Unknown property 'font-size' found, use 'fontSize' instead         react/no-unknown-property
```
I don't think this is preact, I think this is eslint

zachmargolis: solution: 8583bdb3dc66b07d25d9f5c7ff66d2c376ac260f

src/daily-dropoffs-report.tsx
zachmargolis: solution: a3f3a9e50a3f7d020ee8fa6ad1e54d6f27b48d78, h/t @aduth and https://github.com/typescript-eslint/typescript-eslint/issues/2483#issuecomment-687095358

src/daily-dropoffs-report.tsx
```diff
@@ -0,0 +1,376 @@
...
+  );
+}
+
+function DailyDropffsReport(): VNode {
+  const ref = useRef(null as HTMLDivElement | null);
+  const [width, setWidth] = useState(undefined as number | undefined);
+  const { setAgencies } = useContext(AgenciesContext);
+  const { start, finish, agency, env } = useContext(ReportFilterContext);
+
+  const { data } = useQuery(`${start.valueOf()}-${finish.valueOf()}`, () =>
```
aduth, line 337: These cache keys are global, so there's a potential for cache reuse between this report and daily auths data fetching. Could add a simple prefix to create a namespace, e.g.

```suggestion
  const { data } = useQuery(`dropoffs/${start.valueOf()}-${finish.valueOf()}`, () =>
```

src/routes/index.tsx
```diff
@@ -1,10 +1,13 @@
 import { VNode } from "preact";
+import DailyAuthsReport from "../daily-auths-report";
+import DailyDropffsReport from "../daily-dropoffs-report";
 import { Router } from "../router";
 import HomeRoute from "./home-route";
 import ReportRoute from "./report-route";
```
aduth, line 6: I'd probably rename this to reflect the fact it's now a higher-order function, not a component.

e.g. `createReportRoute` or `createReportRouteComponent`

src/table.tsx
```diff
@@ -19,21 +19,25 @@ function Table({ data, numberFormatter = String }: TableProps): VNode {
       <table className="usa-table usa-table--compact">
         <thead>
           <tr>
-            {header.map((head) => (
-              <th>{head}</th>
-            ))}
+            {header.map((head) =>
+              typeof head === "object" && head.type === "th" ? head : <th>{head}</th>
```
aduth, line 23: Another Preact difference: In React we could use [`isValidElement`](https://reactjs.org/docs/react-api.html#isvalidelement) 😅

src/routes/index.tsx
zachmargolis: done f4eba939cc5d4caa9f6c2b3da487401da29b6020

src/table.tsx
zachmargolis: is it time to switch? i don't feel strongly about either react or preact, I just like the sound of "smaller"

aduth: > is it time to switch? i don't feel strongly about either react or preact, I just like the sound of "smaller"

This one's a pretty minor thing, I'm not too concerned. It's a pretty rare use-case and in a sense I have some appreciation for Preact's leanness. Personally I'd be fine to stay on Preact, and I think it could be a helpful stress test if we'd want to leverage it in other settings like [design system](https://github.com/uswds/uswds/issues/3445#issuecomment-674060672) [custom elements](https://preactjs.com/guide/v10/web-components/). The point of being "smaller" has some irony though, because at the same time we pull in [_enormous_](https://bundlephobia.com/package/@observablehq/plot) D3 libraries 😄

zachmargolis:
This is ready for review!

src/routes/index.tsx
```diff
@@ -1,10 +1,13 @@
...
+import DailyDropffsReport from "../daily-dropoffs-report";
 import { Router } from "../router";
 import HomeRoute from "./home-route";
-import ReportRoute from "./report-route";
+import createReportRoute from "./report-route";
 
 export const ROUTES = {
-  "/daily-auths-report/": ReportRoute,
+  "/daily-auths-report/": createReportRoute(DailyAuthsReport, "Daily Auths Report"),
+  "/daily-dropoffs-report/": createReportRoute(DailyDropffsReport, "Daily Dropoffs Report"),
```
aduth, line 10: We got rid of the automated static path generation, so we'll need to add this:

https://github.com/18F/identity-reporting/blob/b871fd559e5a410b0df56e22a69eea33c795e593/bin/generate-routes.js#L5

src/daily-auths-report.tsx
```diff
@@ -279,29 +281,15 @@ function DailyAuthsReport(): VNode {
...
-          </p>
-          <p>
-            <strong>Counting: </strong>
-            This report displays the total number of authentications, so one user authenticating
-            twice will count twice. It does not de-duplicate users or provide unique auths.
-          </p>
-        </div>
-      </div>
+      <Accordion id="how-is-it-measured" title="How is this measured?">
+        <Markdown
```
aduth, line 285: Is there a reason we want this Markdown component, vs. just rendering the markup?

src/accordion.tsx
```diff
@@ -0,0 +1,29 @@
+import { ComponentChildren, VNode } from "preact";
+
+interface AccordionProps {
+  id: string;
+  title: string | VNode;
+  children?: ComponentChildren;
+}
+
+function Accordion({ id, title, children }: AccordionProps): VNode {
```
aduth, line 9: Since the `id` isn't used externally, we could generate it internal to the component?

We use this in the IdP, for example: https://github.com/18F/identity-idp/blob/main/app/javascript/packages/document-capture/hooks/use-instance-id.js

src/daily-dropoffs-report.tsx
```diff
@@ -0,0 +1,392 @@
...
+import { Axis as D3Axis, axisBottom, axisLeft } from "d3-axis";
+import { schemeCategory10 } from "d3-scale-chromatic";
+import Markdown from "preact-markdown";
+import { path as reportPath } from "./report";
+import { ReportFilterContext } from "./context/report-filter-context";
+import Table, { TableData } from "./table";
+import { AgenciesContext } from "./context/agencies-context";
+import Accordion from "./accordion";
+
+// enum Mode {
```
aduth, line 20: Should we remove the commented code?

src/daily-dropoffs-report.tsx
```diff
@@ -0,0 +1,392 @@
...
+  rotateLabels?: boolean;
+  className?: string;
+}): VNode {
+  const ref = useRef(null as SVGGElement | null);
+
+  useEffect(() => {
+    if (ref.current) {
+      select(ref.current).call(axis).classed("rotate-labels", !!rotateLabels);
+    }
+  });
```
aduth, line 241: We should include dependencies here, even if it's just an empty array. Otherwise, `useEffect` is called on every render.

```suggestion
  }, [axis, rotateLabels]);
```

src/daily-dropoffs-report.tsx
```diff
@@ -0,0 +1,392 @@
...
+
+  return <g ref={ref} className={className} transform={transform} />;
+}
+
+function LineChart({
```
aduth, line 246: Can we pull out some of these components to separate files?

src/daily-dropoffs-report.tsx
```diff
@@ -0,0 +1,392 @@
...
+    if (!data) {
+      return;
+    }
+
+    const allAgencies = Array.from(new Set(data.map((d) => d.agency)))
+      .filter((x) => !!x)
+      .sort();
+
+    setAgencies(allAgencies);
+  }, [data]);
```
aduth, line 356: If this is the same as the other report, we could create a custom hook, like `useAgencies(data);`

src/daily-dropoffs-report.tsx
```diff
@@ -0,0 +1,392 @@
...
+
+  useEffect(() => {
+    const listener = () => setWidth(ref.current?.offsetWidth);
+
+    listener();
+
+    window.addEventListener("resize", listener);
+    return () => window.removeEventListener("resize", listener);
+  });
```
aduth, line 365: Same: `useResizeListener(listener)`.

aduth (approved):

src/routes/index.tsx
zachmargolis: oh good catch, thank you! added in 69d1de3141ad9050b893c348981f5bc47d79967b

src/daily-auths-report.tsx
zachmargolis: I felt lazy about writing HTML for prose, I don't feel that strongly about it, I can go back to HTML markup if you'd prefer

src/accordion.tsx
zachmargolis: I like it! Will copy

src/daily-dropoffs-report.tsx
zachmargolis: I had one last boost of energy and added in the feature in eef1c5e634b9ac7f5e898175f192ad3bfe45faec

src/daily-dropoffs-report.tsx
zachmargolis: So this LineChart is very tied to the structure of this particular report, so I'd pull it out as DropOffReportLineChart ?

src/daily-dropoffs-report.tsx
zachmargolis: good point... will do

zachmargolis: got the commit message wrong but extracted in 3bf874d9ded181dedbf4ae54486f364b583ed9c1

src/daily-dropoffs-report.tsx
zachmargolis: 3bf874d9ded181dedbf4ae54486f364b583ed9c1

src/daily-dropoffs-report.tsx
zachmargolis: Ok so... this had a lot of circular dependencies... may have to separate this report's data model out into its own file so it can be shared.. will take a look tomorrow

src/accordion.tsx
zachmargolis: b5ab799edb8dbf80171d4e08e11e56196f452269, 1af1cb5ad44b3f9bdf88e2758f1c5054be9eea42

src/daily-dropoffs-report.tsx
zachmargolis: I have some ideas for breaking up the current report classes into separate data model vs rendering components... going to do that as a separate PR to minimize the noise

aduth: Ah, yeah, the original comment was more in the spirit of "if this is just a generic component to render a line chart", toward an ideal of one-component-per-file. And if it's not a generic component, could it be?

zachmargolis: If Plot's line chart supported ordinal scales like this we could use that instead.... if we add a second line chart like this I could definitely make it more generic, but since it's only one of one, I think it works?

src/hooks/resize-listener.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect, Ref } from "preact/hooks";
+
+function useResizeListener(ref: Ref<HTMLElement | null>, listener: () => void): void {
```
aduth, line 3: The `ref` argument is unused. Can we drop it?

zachmargolis: Whoops, good catch! will make another PR to remove

zachmargolis: https://github.com/18F/identity-reporting/pull/35

src/daily-dropoffs-report.tsx
```diff
@@ -0,0 +1,406 @@
...
+          <g className="dots">
+            {toStepCounts(highlightedRow, funnelMode).map(
+              ({ step, count, percentOfFirst }, idx) => (
+                <>
+                  <circle cx={x(step)} cy={y(count)} r="3" fill={color(highlightedRow.issuer)} />
+                  <text x={x(step)} y={y(count)} font-size="12" dx="3" dy="-3">
+                    <tspan x={x(step)}>{formatWithCommas(count)}</tspan>
+                    {idx > 0 && (
+                      <tspan x={x(step)} dy="1.2em">
+                        (${formatAsPercent(percentOfFirst)})
```
aduth, line 339: Dollar sign isn't necessary in JSX, and is interpolated literally:

image

zachmargolis: 🤦  removed in 137d2b92b5e939981d0789585eeef45ef08dbf2f

PR #32, opened 2021-09-24
Rename ALL THE THINGS! \o/

zachmargolis:
Based on https://github.com/18F/identity-reporting/pull/31#discussion_r715220440

There's no functional change (I hope) from #31 but this touches basically every file in the project so it seemed like a good candidate for a separate PR? 😬 

Could do this more piecemeal if that would make for a better review


.eslintrc.json
```diff
@@ -23,6 +23,7 @@
     "import/no-extraneous-dependencies": ["error", { "devDependencies": ["**/*.test.*", "./*"] }],
     "jsx-a11y/label-has-associated-control": ["error", { "assert": "either" }],
     "no-shadow": "off",
+    "no-unused-vars": "off",
```
zachmargolis, line 26: without this, eslint complained about "unused var Step" ... which is most definitely used, there's a `@typescript/no-unused-vars` we already have enabled as the workaround

src/reports/path.test.ts
```diff
@@ -1,5 +1,5 @@
 import { expect } from "chai";
-import { path } from "./report";
+import { path } from "./path";
```
zachmargolis, line 2: unsure if `reports/path` makes sense as a file? 🤷

src/components/daily-dropoffs-line-chart.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { line as d3Line } from "d3-shape";
+import { formatWithCommas, formatAsPercent } from "../formats";
+import {
+  DailyDropoffsRow,
+  FunnelMode,
+  funnelSteps,
+  StepCount,
+  stepToTitle,
+  toStepCounts,
+} from "../models/daily-dropoffs-report-data";
```
zachmargolis, line 15: originally I had `models/daily-dropoffs-report` and `reports/daily-dropoffs-report` as the two files, but I remembered I hate doing things that way because it makes it hard to open the right file in my editor

src/formats.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const formatWithCommas = format(",");
+
+const formatSIPrefix = format(".2s");
+export const formatSIDropTrailingZeroes = (d: number): string =>
+  formatSIPrefix(d).replace(/\.0+/, "");
+
+export const formatAsPercent = format(".0%");
+
+export const yearMonthDayParse = utcParse("%Y-%m-%d") as (s: string) => Date;
```
zachmargolis, line 14: I think this file is arguably net-new, unsure if it's useful? there were some similar declarations in a few places.. maybe not worth the DRY? 🤷

src/routes/report-route.tsx
```diff
@@ -1,14 +1,16 @@
 import { VNode } from "preact";
 import { utcParse } from "d3-time-format";
 import { utcWeek } from "d3-time";
-import { AgenciesContextProvider } from "../context/agencies-context";
+import { AgenciesContextProvider } from "../contexts/agencies-context";
```
zachmargolis, line 4: the other directories (components, hooks) are plural, so I updated contexts to be plural?

src/formats.ts
aduth: I like the idea of consolidating these, though a bit wary of ad-hoc files. I think it's probably fine, but maybe also consider:

- Could some / all of these be implemented as Preact components? e.g. `const Percent = ({ number }) => format(".0%");`
- Maybe [NPM workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces), to give some structure to ad-hocness?
- Constants / configuration file including desired formats?

src/reports/path.test.ts
aduth: > unsure if `reports/path` makes sense as a file? 🤷

Is a bit confusing to me, to be honest. Maybe keep in root path, or somehow incorporate into `models`, since that's where it's used? What else would a `reports` folder contain?

src/routes/report-route.tsx
aduth: > the other directories (components, hooks) are plural, so I updated contexts to be plural?

I haven't often seen contexts in the plural form, but I think it makes sense even in considering it as a grouping of individual context providers, and for consistency with the other folders.

aduth (approved):

src/reports/path.test.ts
zachmargolis: I'll move it into models, `models/api-path`?

src/formats.ts
zachmargolis: I think some of them could be components for things like the table, but we use the functions directly in a lot of places so we'd still need to expose them?

src/reports/path.test.ts
aduth: > I'll move it into models, `models/api-path`?

It could work. Kinda muddies the distinction between a proper "model" and utility functions common to those implementations. Maybe a subfolder? Or in an `index.js` ?

A different approach could be to adapt what's implemented here as models as custom hooks, e.g. `useDailyAuthsReportData` and `useDropoffsReportData`, at which point maybe `useReportPath` can be a common hook to compose into those?

src/components/daily-dropoffs-line-chart.tsx
```diff
@@ -0,0 +1,112 @@
...
+  width = 400,
+  height = 400,
+}: {
+  data: DailyDropoffsRow[];
+  funnelMode: FunnelMode;
+  color: (issuer: string) => string;
+  width?: number;
+  height?: number;
+}): VNode {
+  const ref = useRef(null as SVGSVGElement | null);
```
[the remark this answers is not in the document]
zachmargolis: nope! removed 94c3226be6fbb02e0490a4c916c498e5a7865d5f

src/reports/path.test.ts
zachmargolis: went with `models/api-path` for now 02593c568a244dfddfb2d0d899e5f13b318d80b5, I didn't have a good insight into the hooks but happy to move this around later

PR #36, opened 2021-09-29
Switch dropoff funnel terminology

zachmargolis:
I got these wrong, so this corrects that, see [notes from grace coworking](https://docs.google.com/document/d/1zGFmLGZpzuJ6ld71V0sMhIpiF01qNAg1thMvZrGkAG0/edit#)

  | Old Name | New Name | Defintion                   |
  | -------- | -------- | --------------------------- |
  | overall  | blanket  | starts at welcome step      |
  | blanket  | actual   | starts at image submit step |

I also figured the definitions could live in the form instead of the explanation accordion:

| before | after |
| --- | --- |
| Screen Shot 2021-09-29 at 11 06 07 AM | Screen Shot 2021-09-29 at 11 02 25 AM|

### DIFF
--- a/src/components/report-filter-controls.tsx
+++ b/src/components/report-filter-controls.tsx
@@ -154,19 +154,6 @@ function ReportFilterControls({ controls }: ReportFilterControlsProps): VNode {
                   <div className="usa-radio">
                     <input
                       type="radio"
-                      id="funnel-mode-overall"
-                      name="funnelMode"
-                      value={FunnelMode.OVERALL}
-                      checked={funnelMode === FunnelMode.OVERALL}
-                      className="usa-radio__input"
-                    />
-                    <label htmlFor="funnel-mode-overall" className="usa-label usa-radio__label">
-                      Overall
-                    </label>
-                  </div>
-                  <div className="usa-radio">
-                    <input
-                      type="radio"
                       id="funnel-mode-blanket"
                       name="funnelMode"
                       value={FunnelMode.BLANKET}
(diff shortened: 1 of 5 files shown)


src/components/report-filter-controls.tsx
```diff
@@ -154,28 +154,32 @@ function ReportFilterControls({ controls }: ReportFilterControlsProps): VNode {
...
+                      value={FunnelMode.BLANKET}
+                      checked={funnelMode === FunnelMode.BLANKET}
                       className="usa-radio__input"
                     />
-                    <label htmlFor="funnel-mode-overall" className="usa-label usa-radio__label">
-                      Overall
+                    <label htmlFor="funnel-mode-blanket" className="usa-label usa-radio__label">
+                      Blanket
                     </label>
+                    <span className="margin-left-1">The funnel starts at the welcome step</span>
```
aduth, line 166: Might be a good use-case for an `aria-describedby` of `input` (in combination with the label)?

https://www.w3.org/TR/WCAG20-TECHS/ARIA1.html

>a label describes the essence of an object, while a description provides more information that the user might need.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute

aduth (approved):

src/components/report-filter-controls.tsx
zachmargolis: sure, added in 8e0b91ea72021b6d29731e0f9c162b04c863e6a8

PR #40, opened 2021-09-30
Add a favicon

zachmargolis:
Screen Shot 2021-09-30 at 3 49 53 PM

@nickttng if you have any other ideas/suggestions I would love to see 'em! My idea was basically

# 📊 + 🛡️

but open to other ideas!


aduth (approved):
From a technical point-of-view, LGTM 👍

zachmargolis:
I'll merge and it we can always drop in a new icon file later if we get one

PR #43, opened 2021-10-20
Add totals row to the Daily Dropoffs Report

zachmargolis:
[Request in Slack](https://gsa-tts.slack.com/archives/C5E7EJWF7/p1634767439004600?thread_ts=1634652839.003400&cid=C5E7EJWF7)

see new footer row

Screen Shot 2021-10-20 at 3 34 14 PM

### DIFF
--- a/src/components/daily-dropoffs-report.tsx
+++ b/src/components/daily-dropoffs-report.tsx
@@ -51,7 +51,7 @@ function tabulate({
       "white" as unknown as number,
     ]);
 
-  const totals = funnelSteps(funnelMode).fill(0);
+  const totals = funnelSteps(funnelMode).map(() => 0);
 
   const body = rows.map((row) => {
     const { agency, issuer, friendly_name: friendlyName } = row;


src/css/style.scss
```diff
@@ -11,6 +11,11 @@
   text-align: right;
 }
 
+.usa-table tfoot td {
+  font-weight: bold;
+  background-color: #dedede;
+}
```
zachmargolis, line 17: As far as I can tell... USWDS does not style `tfoot` at all... I might consider opening a PR that just copies the `thead` ones? 🤷

src/components/table.tsx
```diff
@@ -5,15 +5,34 @@ export type TableRow = (string | number | VNode)[];
...
 
+type NumberFormatter = (n: number) => string;
+
 interface TableProps {
   data: TableData;
-  numberFormatter?: (n: number) => string;
+  numberFormatter?: NumberFormatter;
+}
+
+function renderRow(row: TableRow, numberFormatter: NumberFormatter) {
```
aduth, line 18: Maybe it can be a proper component `TableRow` ?

src/css/style.scss
```diff
@@ -11,6 +11,11 @@
   text-align: right;
 }
 
+.usa-table tfoot td {
+  font-weight: bold;
+  background-color: #dedede;
```
aduth, line 16: Could avoid hard-coding and use the token instead:

```suggestion
  background-color: color('base-lighter');
```

Via: https://design.login.gov/utilities/color/

src/components/daily-dropoffs-report.tsx
```diff
@@ -51,6 +51,8 @@ function tabulate({
       "white" as unknown as number,
     ]);
 
+  const totals = funnelSteps(funnelMode).map(() => 0);
```
aduth, line 54: Good use case for `Array#fill`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill

src/components/table.tsx
zachmargolis: updated in  a185371dc8eb06e4b8af37f46dabb2a9f0f0de90

PR #48, opened 2021-12-14
include end date in range

mitchellhenke:
Currently, selecting the range "December 12th, 2021" - "December 13th, 2021" in the UI results in querying only December 12th because `utcDays(start, finish, 1)` returns a date range excluding the final date.

This patch updates the behavior to include the end date in the queries using the recommendation found in a d3-time issue [here](https://github.com/d3/d3-array/issues/90#issuecomment-642761443)


zachmargolis:
I just noticed this PR now, going to close it out since we have https://github.com/18F/identity-reporting/pull/49 now, using `utcDays.offset()` instead of the `.concat()` trick

PR #49, opened 2022-01-18
Fix last day not being included in report (LG-5513)

nathanberg:
Specs were also updated.

Core issue is that the "finish" date would be set to the beginning of the day, thus not include information that persists throughout the rest of the day.

Solution was to set the timestamp to the very end of the day - this ensures that we don't accidentally add a day to automated reports.

Originally something like 01/01/2020-01/03/2020 would only include 01/01/2020 @ 12:00AM - 01/03/2020 @ 12:00AM

This fix changes this behavior to report 01/01/2020 @ 12:00AM - 01/03/2020 @ 11:59PM as an example.

Specs had to be adjusted to recognize the change.

### DIFF
--- a/src/models/daily-auths-report-data.ts
+++ b/src/models/daily-auths-report-data.ts
@@ -1,4 +1,4 @@
-import { utcDays, timeDay } from "d3-time";
+import { utcDays, utcDay } from "d3-time";
 import { path as reportPath } from "./api-path";
 
 interface Result {
@@ -49,7 +49,7 @@ function loadData(
   fetch = window.fetch
 ): Promise<ProcessedResult[]> {
   return Promise.all(
-    utcDays(start, timeDay.ceil(finish), 1).map((date) => {
+    utcDays(start, utcDay.offset(finish, 1), 1).map((date) => {
       const path = reportPath({ reportName: "daily-auths-report", date, env });
       return fetch(path).then((response) =>
         response.status === 200 ? response.json() : { results: [] }
--- a/src/models/daily-auths-report-data.ts
+++ b/src/models/daily-auths-report-data.ts
@@ -1,4 +1,4 @@
-import { utcDays } from "d3-time";
+import { utcDays, timeDay } from "d3-time";
 import { path as reportPath } from "./api-path";
 
 interface Result {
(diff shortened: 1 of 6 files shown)


src/models/daily-dropoffs-report-data.ts
```diff
@@ -158,6 +158,8 @@ function loadData(
   env: string,
   fetch = window.fetch
 ): Promise<DailyDropoffsRow[]> {
+  // Set the "finish" to the ending of the day
+  finish.setUTCHours(23, 59, 59, 999);
```
zachmargolis, line 162: 1. looks like this mutates the parameter to the function? I think that's a pattern we should try to avoid, can we make a copy of the param if we're going to mutate it?

2. Is there a way we can use `utcDay.ceil()` or some similar helper function to advance this?

nathanberg: I've used the timeDay.Ceil() function to accomplish Item 2. Think it's still necessary to make a copy of the param while utilizing the .ceil() function?

zachmargolis: yup! AFAIK `.ceil` should make a new instance

zachmargolis:
I checked this out locally (looks like we're having some CORS issues on the data side we can address separately), but I wanted to point out a few bugs on the Daily Auths Report

- adding this new date change in the data request layer means the daily auths report is receiving additional data and rendering an extra bar that gets cut off
- the UI shows  08/08/2021-08/15/2021 because it defaults to "a week", I think if our weeks are now inclusive, we should probably update that logic to be 6 days aka 08/08-08/14
- at the end of the day, I think no matter what we should be trying to render 7 columns for a week, so defaulting to 8 here needs to be cleaned up

| before | after |
| --- |--- |
| Screen Shot 2022-01-18 at 10 55 29 AM | Screen Shot 2022-01-18 at 10 55 33 AM |

src/models/daily-auths-report-data.ts
```diff
@@ -1,4 +1,4 @@
-import { utcDays } from "d3-time";
+import { utcDays, timeDay } from "d3-time";
```
zachmargolis, line 1: I think `timeDay` may be local timezone, but to keep things consistent we should use `utcDay`
```suggestion
import { utcDays, utcDay } from "d3-time";
```

nathanberg:
Fixed the graphical issue :) @zachmargolis

zachmargolis (approved):
LGTM, checked it out locally and this makes sense to me!

PR #58, opened 2022-06-17
Bump eslint packages, fix lint warnings

zachmargolis:
Based on https://github.com/18F/identity-reporting/pull/57#discussion_r900452102

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,11 +1,8 @@
 {
   "root": true,
   "parser": "@typescript-eslint/parser",
-  "plugins": ["@18f/eslint-plugin-identity", "@typescript-eslint"],
-  "extends": [
-    "plugin:@18f/eslint-plugin-identity/recommended",
-    "plugin:@typescript-eslint/recommended"
-  ],
+  "extends": ["plugin:@18f/eslint-plugin-identity/recommended"],
+  "plugins": ["@18f/eslint-plugin-identity"],
   "parserOptions": {
     "sourceType": "module",
     "ecmaVersion": 2021
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,5 @@
 {
   "root": true,
-  "parser": "@typescript-eslint/parser",
   "extends": ["plugin:@18f/eslint-plugin-identity/recommended"],
   "plugins": ["@18f/eslint-plugin-identity"],
   "parserOptions": {
@@ -12,23 +11,14 @@
     "browser": true
   },
   "rules": {
-    "@typescript-eslint/explicit-module-boundary-types": "error",
-    "@typescript-eslint/no-explicit-any": "error",
-    "@typescript-eslint/no-non-null-assertion": "error",
-    "@typescript-eslint/no-shadow": "error",
-    "@typescript-eslint/no-unused-vars": "error",
-    "import/no-extraneous-dependencies": ["error", { "devDependencies": ["**/*.test.*", "./*"] }],
     "jsx-a11y/label-has-associated-control": ["error", { "assert": "either" }],
-    "no-shadow": "off",
-    "no-unused-vars": "off",
-    "react/jsx-filename-extension": ["error", { "extensions": [".tsx"] }],
+    "import/no-extraneous-dependencies": ["error", { "devDependencies": ["**/*.test.*", "./*"] }],
     "react/no-unknown-property": [
       "error",
       { "ignore": ["font-size", "stroke-width", "text-anchor"] }
     ]
   },
   "settings": {
-    "import/resolver": "eslint-import-resolver-typescript",
     "react": {
       "pragma": "h",
       "fragment": "Fragment",
--- a/src/components/header.tsx
+++ b/src/components/header.tsx
@@ -1,7 +1,9 @@
 import { VNode } from "preact";
-import logoURL from "../../node_modules/identity-style-guide/dist/assets/img/login-gov-logo.svg";
-import closeURL from "../../node_modules/identity-style-guide/dist/assets/img/close.svg";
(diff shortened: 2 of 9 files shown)


src/models/daily-auths-report-data.test.ts
```diff
@@ -25,7 +25,7 @@ describe("DailyAuthsReportData", () => {
         yearMonthDayParse("2021-01-01"),
         yearMonthDayParse("2021-01-03"),
         "local",
-        fetch
+        fetch as typeof window.fetch
```
zachmargolis, line 28: fetch-mock's types seemed to require a `URL` but we were passing in `string` so this was a quick way to tell it to be quiet 😬 

I tried updating to send a URL in the code, but then that threw with a few "invalid URL" errors

.eslintrc.json
```diff
@@ -1,11 +1,8 @@
 {
   "root": true,
   "parser": "@typescript-eslint/parser",
-  "plugins": ["@18f/eslint-plugin-identity", "@typescript-eslint"],
-  "extends": [
-    "plugin:@18f/eslint-plugin-identity/recommended",
-    "plugin:@typescript-eslint/recommended"
-  ],
+  "extends": ["plugin:@18f/eslint-plugin-identity/recommended"],
+  "plugins": ["@18f/eslint-plugin-identity"],
```
aduth, line 5: Can we now remove any / all of the explicit `rules` defined below?

src/components/header.tsx
```diff
@@ -1,7 +1,9 @@
 import { VNode } from "preact";
+import { getFullPath, Link } from "../router";
+// eslint-disable-next-line import/no-relative-packages
 import logoURL from "../../node_modules/identity-style-guide/dist/assets/img/login-gov-logo.svg";
```
aduth, line 4: Out of curiosity, would it work to import it from the package name base?

```suggestion
import logoURL from "identity-style-guide/dist/assets/img/login-gov-logo.svg";
```

src/models/daily-auths-report-data.test.ts
aduth: Yeah, I think the `fetch-mock` types are incorrect because the mocked function signature doesn't allow for `URL` objects, which are valid in `fetch` global.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e7c664fcae5f8116097f01bb05d7b2d6d2cb3dde/types/fetch-mock/index.d.ts#L626

(FYI, was there a reason we use DefinitelyTyped for this package's types, rather than the [package's first-party types](https://github.com/wheresrhys/fetch-mock/tree/master/types)? From some initial testing, it won't help with this specific problem though)

src/components/table.tsx
```diff
@@ -2,7 +2,8 @@ import { VNode, render } from "preact";
 import { csvFormatValue } from "d3-dsv";
 import Icon from "./icon";
 
-export type TableCell = string | number | VNode;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
aduth, line 5: Per prior point, if we can just remove this rule override in the ESLint config, we wouldn't need to disable the rule.

aduth (approved):
👍

src/models/daily-auths-report-data.test.ts
aduth: Looks like we can shrink to at least this:

```json
{
  "root": true,
  "extends": ["plugin:@18f/eslint-plugin-identity/recommended"],
  "plugins": ["@18f/eslint-plugin-identity"],
  "parserOptions": {
    "sourceType": "module",
    "ecmaVersion": 2021
  },
  "env": {
    "node": true,
    "browser": true
  },
  "rules": {
    "no-shadow": "off",
    "react/no-unknown-property": [
      "error",
      { "ignore": ["font-size", "stroke-width", "text-anchor"] }
    ]
  },
  "settings": {
    "react": {
      "pragma": "h",
      "fragment": "Fragment",
      "version": "16.2"
    }
  },
  "ignorePatterns": ["_site/**"]
}
```

.eslintrc.json
aduth: Looks like we can shrink to at least this:

```json
{
  "root": true,
  "extends": ["plugin:@18f/eslint-plugin-identity/recommended"],
  "plugins": ["@18f/eslint-plugin-identity"],
  "parserOptions": {
    "sourceType": "module",
    "ecmaVersion": 2021
  },
  "env": {
    "node": true,
    "browser": true
  },
  "rules": {
    "no-shadow": "off",
    "react/no-unknown-property": [
      "error",
      { "ignore": ["font-size", "stroke-width", "text-anchor"] }
    ]
  },
  "settings": {
    "react": {
      "pragma": "h",
      "fragment": "Fragment",
      "version": "16.2"
    }
  },
  "ignorePatterns": ["_site/**"]
}
```

zachmargolis: love it, 6ad382d7d95ae7e7d917cf05b570d54e7fc53c89

src/components/header.tsx
zachmargolis: vite was upset:
```
1:26:58 PM [vite] Internal server error: Missing "./dist/assets/img/login-gov-logo.svg" export in "identity-style-guide" package
  Plugin: vite:import-analysis
```

src/components/table.tsx
zachmargolis: gone: 5fe1f51aa670f485f41791c65368114dad6da057

src/models/daily-auths-report-data.test.ts
zachmargolis: I think it was just habit of mine to install them that way... removed in d970c2eff25e33b29d34a8e475ce75f6a53ee4e6

PR #59, opened 2022-06-17
Separate issuer and friendly name in daily auths CSV

zachmargolis:
This matches the style in the dropoffs report better, and is clearer data for actual processing

| before | after |
| --- | --- |
| Screen Shot 2022-06-17 at 1 20 53 PM | Screen Shot 2022-06-17 at 1 20 45 PM |


theabrad (approved):
LGTM

PR #60, opened 2022-06-17
Fix kebab casing

zachmargolis:
- Add regression specs


theabrad (approved):
Screen Shot 2022-06-17 at 5 07 39 PM

PR #61, opened 2022-06-22
Move aggregation behavior for "daily dropoffs" outside of "loadData"

zachmargolis:
**Why**: That way, loadData exposes more raw data, so we can process
it differently if needed

I was starting work on LG-6540, and this was causing a bunch of data
errors, because I was expecting de-aggregated data

### DIFF
--- a/src/components/daily-dropoffs-report.tsx
+++ b/src/components/daily-dropoffs-report.tsx
@@ -17,6 +17,7 @@ import {
   loadData,
   toStepCounts,
   aggregateAll,
+  aggregate,
 } from "../models/daily-dropoffs-report-data";
 import { formatAsPercent, formatWithCommas, yearMonthDayFormat } from "../formats";
 
@@ -124,7 +125,7 @@ function DailyDropffsReport(): VNode {
   useResizeListener(() => setWidth(ref.current?.offsetWidth));
   useAgencies(data);
 
-  const nonNullData = data || [];
+  const nonNullData = aggregate(data || []);
   const filteredData = (byAgency ? nonNullData : aggregateAll(nonNullData)).filter(
     (d) => !agency || d.agency === agency
   );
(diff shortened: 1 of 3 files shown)


src/models/daily-dropoffs-report-data.test.ts
```diff
@@ -164,37 +165,52 @@ describe("DailyDropoffsReportData", () => {
   describe("#loadData", () => {
     const yearMonthDayParse = utcParse("%Y-%m-%d") as (s: string) => Date;
 
-    it("combines data across separate fetch requests", () => {
+    it("concatenates data across separate fetch requests", () => {
```
zachmargolis, line 168: luckily we already have tests for the aggregate behavior on its own: https://github.com/18F/identity-reporting/blob/2d014b9de3417974079e34c37aee2d3e43e22cb7/src/models/daily-dropoffs-report-data.test.ts#L15

src/components/daily-dropoffs-report.tsx
```diff
@@ -124,7 +125,7 @@ function DailyDropffsReport(): VNode {
   useResizeListener(() => setWidth(ref.current?.offsetWidth));
   useAgencies(data);
 
-  const nonNullData = data || [];
+  const nonNullData = aggregate(data || []);
```
aduth, line 128: Just to confirm: The other places we use `data` or call `loadData` don't need to be using the aggregated result?

https://github.com/18F/identity-reporting/blob/0afee59a170b0ac1aebe2a54c7dba21976bdfc45/src/components/daily-dropoffs-report.tsx#L126

https://github.com/18F/identity-reporting/blob/0afee59a170b0ac1aebe2a54c7dba21976bdfc45/src/components/daily-auths-report.tsx#L233-L235

aduth (approved):
👍

src/components/daily-dropoffs-report.tsx
zachmargolis: Correct, it shouldn't matter, `useAgencies` will grab the unique agencies, so it doesn't matter if it's been grouped by issuer or not (since there can be multiple issuers per agency anyways)

PR #62, opened 2022-06-23
Add week-over-week proofing (LG-6540)

zachmargolis:
This report doesn't have a table/downloadable part yet, figured might make for a good follow-up PR instead of one super giant one

Also, so this PR makes super heavy use of Plot, and I haven't figured out a great way to test that, other than to look at the charts 😬 

Here are some of the different modes. I could only get the "average" trendline to work with percentages... would be nice to get it fixed later

| | percent | count |
| -- | --- | ---- |
| weekly |  Screen Shot 2022-06-23 at 4 28 39 PM |  Screen Shot 2022-06-23 at 4 28 30 PM |
| daily | Screen Shot 2022-06-23 at 4 28 35 PM  | Screen Shot 2022-06-23 at 4 28 45 PM |

And here is a screenshot of some earlier ideas I was noodling on, but I think it's too noisy and not clear enough as-is for what we'd want to show (also would need to be stacked slightly differently, but worth considering!)
| screenshot |
| --- |
| Screen Shot 2022-06-22 at 4 29 10 PM |


src/components/report-filter-controls.tsx
```diff
@@ -21,15 +22,27 @@ enum Control {
   SCALE = "scale",
   AGENCY = "agency",
   BY_AGENCY = "by_agency",
+  TIME_BUCKET = "time_bucket",
```
zachmargolis, line 25: Another name for time bucket would be like, "period" but I wasn't sure if that was clearer or not 🤷 . Open to feedback on what to call this!

src/components/report-filter-controls.tsx
```diff
@@ -106,6 +119,37 @@ function ReportFilterControls({ controls }: ReportFilterControlsProps): VNode {
...
+                      id="time-bucket-day"
+                      name="timeBucket"
+                      value="day"
+                      checked={timeBucket === TimeBucket.DAY}
+                      className="usa-radio__input"
+                    />
+                    <label htmlFor="time-bucket-day" className="usa-label usa-radio__label">
+                      Day
+                    </label>
+                  </div>
```
zachmargolis, line 137: Future refactor idea: make a component for these radio controls, it's a lot of markup and manual ID construction for only 2 options

src/components/proofing-over-time-report.tsx
```diff
@@ -0,0 +1,188 @@
...
+    if (byAgency) {
+      if (agency) {
+        return "friendlyName";
+      }
+      return "agency";
+    }
+  })();
+
+  const filter = (d: StepCountEntry) =>
+    d.step === Step.VERIFIED && (!agency || d.agency === agency);
```
aduth, line 81: Trying to follow the logic here: Is the step condition something which would be better to filter out earlier, like in `flatten`'s `flatMap` ? Just observing with some breakpoints that with larger spans of time (which I expect would be a common use-case), there's tens of thousands of data rows. Dunno if this could help prevent any potential performance issues.

src/components/proofing-over-time-report.tsx
```diff
@@ -0,0 +1,188 @@
...
+    loadData(start, finish, env)
+  );
+
+  useResizeListener(() => setWidth(ref.current?.offsetWidth));
+  useAgencies(data);
+
+  const flatSteps = flatten({ data: data || [], funnelMode });
+
+  const lineDeterminer = (() => {
+    if (byAgency) {
```
aduth, line 72: With `byAgency` filtering, could we have some label shown when hovering a line, similar to how it works in the other reports?

aduth (approved):
This is a really useful report! 👍

src/components/proofing-over-time-report.tsx
zachmargolis: That would probably help. I think I was keeping the data around in case we found a way to render the various steps like in that exploration screenshot

src/components/proofing-over-time-report.tsx
aduth: >My instinct is to save that for a second pass instead of waiting on it? How strongly do you feel?

Not that strongly, especially since `byAgency` isn't really exposed at the moment.

PR #63, opened 2022-06-28
WIP: add vertical lines for releases

zachmargolis:
Creating PR as a demo/record.

I don't think that the values seem to change enough that showing these really helps, but happy to discuss later

Screen Shot 2022-06-28 at 2 15 03 PM

### DIFF
--- a/src/github.ts
+++ b/src/github.ts
@@ -1,0 +1,23 @@
+interface Release {
+  name: string;
+  createdAt: Date;
+  tagName: string;
+}
+
+interface ApiRelease {
+  name: string;
+  created_at: string;
+  tag_name: string;
+}
+
+function loadReleases({ owner, repo }: { owner: string; repo: string }): Promise<Release[]> {
+  return fetch(`https://api.github.com/repos/${owner}/${repo}/releases`)
+    .then((response) => response.json())
+    .then((releases: ApiRelease[]) =>
+      releases.map(({ name, created_at: createdAt, tag_name: tagName }) => {
+        return { name, createdAt: new Date(createdAt), tagName };
+      })
+    );
+}
+
+export { loadReleases };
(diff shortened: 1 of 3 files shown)


src/github.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+function loadReleases({ owner, repo }: { owner: string; repo: string }): Promise<Release[]> {
+  return fetch(`https://api.github.com/repos/${owner}/${repo}/releases`)
+    .then((response) => response.json())
+    .then((releases: ApiRelease[]) =>
+      releases.map(({ name, created_at: createdAt, tag_name: tagName }) => {
+        return { name, createdAt: new Date(createdAt), tagName };
+      })
+    );
```
aduth, line 20: `async` / `await` style?

```suggestion
async function loadReleases({ owner, repo }: { owner: string; repo: string }): Promise<Release[]> {
  const response = await fetch(`https://api.github.com/repos/${owner}/${repo}/releases`);
  const releases: ApiRelease[] = await response.json();
  return releases.map(({ name, created_at: createdAt, tag_name: tagName }) => ({
    name,
    createdAt: new Date(createdAt),
    tagName,
  }));
```

src/github.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+interface ApiRelease {
+  name: string;
+  created_at: string;
+  tag_name: string;
+}
+
+function loadReleases({ owner, repo }: { owner: string; repo: string }): Promise<Release[]> {
+  return fetch(`https://api.github.com/repos/${owner}/${repo}/releases`)
```
aduth, line 14: Interesting [the docs](https://docs.github.com/en/rest/releases/releases#list-releases) don't specify it, but I suppose it's safe to assume these are ordered by most recent creation date?

aduth: With longer spans of time, would we have to deal with pagination?

aduth (commented):
Seems like it could be useful if we're expecting numbers to go up and down corresponding to changes shipped through releases

PR #66, opened 2022-08-17
Update proofing-over-time table to show verified count

zachmargolis:
**Why**: We care about the number of successes more than
the number of attempts

Highlighted by @porta-antiporta, thanks!


aduth (approved):
👍

PR #70, opened 2022-10-05
Migrate Federalist to Cloud.gov Pages

aduth:
**Why**: [Federalist is becoming Cloud.gov Pages](https://cloud.gov/pages/federalist-migration/), and as such we'll need to update configurations accordingly.

The webhook settings have already been updated, so these changes are updating the remaining code reference to Federalist.

Resources:

- [Slack confirmation of renamed script](https://gsa-tts.slack.com/archives/C1NUUGTT5/p1664896538701049)
- [Related code](https://github.com/cloud-gov/pages-build-container/blob/508bd8f/src/steps/build.py#L186-L200)

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "postbuild": "npm run generate-routes && cp robots.txt _site/",
     "dev": "vite",
     "generate-routes": "node -r esbuild-register bin/generate-routes.js",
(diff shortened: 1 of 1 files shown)


zachmargolis (approved):
LGTM

PR #74, opened 2023-01-19
Re-use ROUTES constant so it's easier to add reports to the nav

zachmargolis:
I'm prepping to add a new report, and this is a small refactor that lets us reduce some boilerplate/repetition across `index.tsx` and `report-routes.tsx`

no visible changes


src/routes/index.tsx
```diff
@@ -4,11 +4,13 @@ import DailyAuthsReport from "../components/daily-auths-report";
...
 import { Router } from "../router";
-import HomeRoute from "./home-route";
-import createReportRoute from "./report-route";
+import createReportRoute, { ReportRoute } from "./report-route";
 import { Scale } from "../contexts/report-filter-context";
+import Page from "../components/page";
+import Header from "../components/header";
 
-export const ROUTES = {
+const ROUTES: Record<string, ReportRoute> = {
```
zachmargolis, line 12: I think originally the goal was to export `ROUTES` so that we could use it in `bin/generate-routes.js` but that's hard to do because SVG importing seems to get in the way

src/routes/index.tsx
```diff
@@ -23,14 +25,18 @@ export const ROUTES = {
...
 export function Routes(): VNode {
   return (
     <Router>
-      {Object.entries(ROUTES).map(([path, Component]) => (
-        <Component path={path} />
+      {Object.entries(ROUTES).map(([path, { title, component: Component }]) => (
+        <div path={path}>
+          <Header path={path} routes={ROUTES} />
+          <Page title={title}>
+            <Component />
```
zachmargolis, line 37: just realized this messes up our query param stuff... `<Component>` needs to be the top level thing routed to

PR #75, opened 2023-01-20
Create ALL_ROUTES constant, define all routes in one place

zachmargolis:
Used by:
- generate-routes (static node script)
- header.tsx (for the nav)
- index.tsx (to map to actual report compontent)

This is round 2 of https://github.com/18F/identity-reporting/pull/74, I opted to create a second file and use TS to require full usage, instead of trying to get cute and only have one file

### DIFF
--- a/src/routes/index.tsx
+++ b/src/routes/index.tsx
@@ -33,7 +33,7 @@ export function Routes(): VNode {
   return (
     <Router>
       {Object.entries(reportRoutes).map(([path, Component]) => (
-        <Component path={path} />
+        <Component path={path as keyof ReportRoutes} />
       ))}
     </Router>
   );
--- a/src/routes/index.tsx
+++ b/src/routes/index.tsx
@@ -5,10 +5,15 @@ import DailyDropffsReport from "../components/daily-dropoffs-report";
 import ProofingOverTimeReport from "../components/proofing-over-time-report";
 import { Router } from "../router";
 import HomeRoute from "./home-route";
-import createReportRoute from "./report-route";
+import createReportRoute, { ReportRoute } from "./report-route";
 import { Scale } from "../contexts/report-filter-context";
+import ALL_ROUTES from "./all";
 
-export const ROUTES = {
+type ReportRoutes = {
+  [Property in keyof typeof ALL_ROUTES]: ReportRoute;
+}
+
+const reportRoutes: ReportRoutes = {
   "/daily-auths-report/": createReportRoute(DailyAuthsReport, {
     title: "Daily Auths Report",
     controls: [Control.IAL],
@@ -29,7 +34,7 @@ export const ROUTES = {
 export function Routes(): VNode {
   return (
     <Router>
-      {Object.entries(ROUTES).map(([path, Component]) => (
+      {Object.entries(reportRoutes).map(([path, Component]) => (
         <Component path={path} />
       ))}
     </Router>
--- a/src/routes/index.tsx
+++ b/src/routes/index.tsx
@@ -12,9 +12,7 @@ import ALL_ROUTES from "./all";
 /**
  * Requires that all keys in ALL_ROUTES have a matching key in this object
  */
-type ReportRoutes = {
-  [Property in keyof typeof ALL_ROUTES]: ReportRoute;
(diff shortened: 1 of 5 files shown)


src/routes/index.tsx
```diff
@@ -5,10 +5,18 @@ import DailyDropffsReport from "../components/daily-dropoffs-report";
...
 
-export const ROUTES = {
+/**
+ * Requires that all keys in ALL_ROUTES have a matching key in this object
+ */
+type ReportRoutes = {
+  [Property in keyof typeof ALL_ROUTES]: ReportRoute;
+}
+
+const reportRoutes: ReportRoutes = {
```
zachmargolis, line 19: I confirmed that this will yell if `reportRoutes` is missing a key that's in `ALL_ROUTES`

src/routes/index.tsx
```diff
@@ -5,20 +5,25 @@ import DailyDropffsReport from "../components/daily-dropoffs-report";
...
 import { Scale } from "../contexts/report-filter-context";
+import ALL_ROUTES from "./all";
 
-export const ROUTES = {
+/**
+ * Requires that all keys in ALL_ROUTES have a matching key in this object
+ */
+type ReportRoutes = {
+  [Property in keyof typeof ALL_ROUTES]: ReportRoute;
+}
```
aduth, line 17: Could express this as a Record perhaps. WDYT?

```suggestion
type ReportRoutes = Record<keyof typeof ALL_ROUTES, ReportRoute>;
```

src/routes/report-route.tsx
```diff
@@ -91,6 +92,8 @@ function createReportRoute(
       reportControls.push(Control.BY_AGENCY);
     }
 
+    const title = (ALL_ROUTES as Record<string, string>)[path];
```
aduth, line 95: Could we avoid the type-casting by assigning the `path` type on `ReportRouteProps` as `keyof typeof ALL_ROUTES` ? Or would that cause some upstream problems for how the component is rendered?

src/routes/index.tsx
```diff
@@ -5,20 +5,25 @@ import DailyDropffsReport from "../components/daily-dropoffs-report";
...
+
+const reportRoutes: ReportRoutes = {
```
aduth, line 19: For future consideration, we may be able to collapse the route constant back in this file if we moved out the static imports which were causing it trouble, using [`React.lazy`](https://reactjs.org/docs/code-splitting.html).

Thinking something like...

```ts
const ROUTES = {
  '/daily-auths-report': {
    title: 'Daily Auths Report',
    component: lazy(() => import('./daily-dropoffs-report')),
  },
};
```

aduth (approved):
LGTM 👍

src/routes/index.tsx
zachmargolis: love it, for some reason I thought records didn't require keys that way

src/routes/index.tsx
zachmargolis: I think we'd have to use [preact-lazy](https://www.npmjs.com/package/preact-lazy) right?

src/routes/report-route.tsx
zachmargolis: It worked! Sort of, but the `Object.entries().map()` lost some fidelity of the keys. PTAL at c1f74d6618489982494659058b8a4030d2f7cc30, it still has a cast, but's more specific?

src/routes/index.tsx
aduth: Oh, right, I forgot we're using Preact here! I'm not super familiar with that specific library, but I imagine it's probably going for the same idea, yeah.

aduth: Though it looks like `preact/compat` has a `lazy` function:

https://preactjs.com/guide/v10/switching-to-preact#suspense

src/routes/report-route.tsx
aduth: Yeah that looks good to me 👍

src/routes/index.tsx
zachmargolis: sounds like a problem for future us 😎

PR #77, opened 2023-01-20
Add Daily/Cumulative Registrations Report (LG-8679)

zachmargolis:
Data relies on https://github.com/18F/identity-idp/pull/7674

I'm waiting for the final OK to merge this, but the resulting report will look something like (redacting numbers for now)

| cumulative | daily |
|----|----|
|  Screen Shot 2023-01-20 at 12 17 59 PM | Screen Shot 2023-01-20 at 12 19 17 PM |

### DIFF
--- a/src/components/daily-registrations-report.tsx
+++ b/src/components/daily-registrations-report.tsx
@@ -2,6 +2,7 @@ import { VNode } from "preact";
 import { useQuery } from "preact-fetching";
 import { useRef, useState, useContext } from "preact/hooks";
 import * as Plot from "@observablehq/plot";
+import Markdown from "preact-markdown";
 import { ReportFilterContext } from "../contexts/report-filter-context";
 import useResizeListener from "../hooks/resize-listener";
 import {
@@ -15,15 +16,8 @@ import PlotComponent from "./plot";
 import { formatSIDropTrailingZeroes, formatWithCommas, yearMonthDayFormat } from "../formats";
 import Table, { TableData } from "./table";
 import Accordion from "./accordion";
-import Markdown from "preact-markdown";
 
-function plot({
-  data,
-  width,
-}: {
-  data: ProcessedRenderableData[];
-  width?: number;
-}): HTMLElement {
+function plot({ data, width }: { data: ProcessedRenderableData[]; width?: number }): HTMLElement {
   return Plot.plot({
     color: {
       legend: true,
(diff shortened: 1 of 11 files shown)


src/components/daily-registrations-report.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+  });
+
+  const windowedData = (data || []).filter(({ date }) => start <= date && date <= finish);
+
+  return (
+    <div ref={ref}>
+      {data && (
+        <PlotComponent
+          plotter={() => plot({ data: filteredData, start, finish, width })}
```
aduth, line 108: Should we bother passing `start` and `finish` if they're unused in the function?

src/components/report-filter-controls.tsx
```diff
@@ -68,6 +70,7 @@ function ReportFilterControls({ controls }: ReportFilterControlsProps): VNode {
   }
 
   return (
+    <div>
```
aduth, line 73: Do we need the extra `<div />`?

src/models/daily-registrations-report-data.ts
```diff
@@ -0,0 +1,91 @@
...
+  type: DataType;
+}
+
+interface DailyRegistrationsReportData {
+  results: Result[];
+
+  /**
+   * ISO8601 string
+   */
+  finish: string;
```
aduth, line 40: What do we use this for?

aduth (approved):
LGTM, but I wasn't able to test. Remind me: How does one test this?

src/components/daily-registrations-report.tsx
zachmargolis: I think at one point I was considering letting those filter the graph instead of having it show all-time, but yes you're right I should remove

src/components/report-filter-controls.tsx
zachmargolis: Technically no... I noticed that this whole form was indented wrong so when prettier fixed it, it created a ton of diff noise (made it harder to see that I just added a single section in this PR), so adding that tag let me clean up the diff 😬 

In hindsight I could have recommended viewing with ?w=1 to hide that, I'll remove it

src/models/daily-registrations-report-data.ts
zachmargolis: We don't, but it's in the response JSON so I figured I'd document it

zachmargolis:
> LGTM, but I wasn't able to test. Remind me: How does one test this?

1. Run the report and generate a JSON file (I'll DM one to you) and put it in a top-level directory called `local` (note the `.gitignore` change)
2. View the report with `?env=local`

aduth (approved):
Tested, LGTM

zachmargolis:
Got the OK to ship this! 🚀

PR #79, opened 2023-01-31
[LG-8794] Add Phone Submit step to dropoff report

orenyk:
Resolves LG-8794

Blocked on https://github.com/18F/identity-idp/pull/7733 before useful data will arrive


zachmargolis (approved):
LGTM but we need to wait until the job is deployed + backfilled before we can merge

PR #82, opened 2023-02-06
Update daily registrations table view: one row per day

zachmargolis:
**Why**: makes for less horizontal scrolling when viewing a longer range of time

| before | after |
| --- | --- |
| Screen Shot 2023-02-06 at 12 53 45 PM | Screen Shot 2023-02-06 at 12 53 51 PM |


aduth (approved):
LGTM

PR #83, opened 2023-02-07
Hide misleading proofing reports

zachmargolis:
**Why**: The graphs are misleading at the moment so this helps us reduce confusion.

See [slack thread for context](https://gsa-tts.slack.com/archives/C01541F40SW/p1675777991411129)

| screenshot |
| --- |
|  Screen Shot 2023-02-07 at 10 08 14 AM |

### DIFF
--- a/src/components/daily-dropoffs-report.tsx
+++ b/src/components/daily-dropoffs-report.tsx
@@ -1,5 +1,5 @@
 import { VNode } from "preact";
-import { useContext, useRef, useState } from "preact/hooks";
+import { useContext, useRef } from "preact/hooks";
 import { useQuery } from "preact-fetching";
 import { scaleLinear, scaleOrdinal } from "d3-scale";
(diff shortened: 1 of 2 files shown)


src/components/daily-dropoffs-report.tsx
```diff
@@ -145,13 +145,15 @@ The data model table can't accurately capture:
...
         />
       </Accordion>
-      <DailyDropoffsLineChart
-        data={filteredData}
-        width={width}
-        color={issuerColor}
-        funnelMode={funnelMode}
-        scale={scale}
-      />
+      {undefined && (
```
zachmargolis, line 148: I would have made this depend on the `extra` flag like in the other report, but doing that, in combination with how this report splits out agencies vs aggregates them, made this whole report too slow.

mitchellhenke (approved):

PR #84, opened 2023-02-09
Re-Add Phone Submit step to dropoff report

orenyk:
This reverts commit 7574503774aa958a65061b7db230d7f514cd3212.


zachmargolis (approved):
LGTM but this report is offline right now so it won't be a visible change

orenyk:
@zachmargolis there were still some issues with the data so I don't think we should add this in the current state. I'm going to close this PR for now, we can always reopen.

PR #86, opened 2023-02-14
Hide Proofing, Not Auths

zachmargolis:
There was a miscommunication leading up to #85, this remedies that by:

- Hiding the Proofing Over Time report
- Brings back the Daily Auths Report


aduth (approved):

zachmargolis:
I triple checked the preview branch for this with @TiffanyAndrews, got the OK, am merging

PR #87, opened 2023-02-21
Bring Back Reports, Annotate Bad Data

zachmargolis:
(First few commits are just reverts, all the new stuff is in the last commit)

When these two reports have date ranges selected that overlap the bad data, we show a banner. And on the "over time" report, we show a big red rectangle highlighting the bad data.

The tables are also affected, but they don't have any warnings as of now.

| dropoffs report | over time report | 
| --- | --- |
| Screen Shot 2023-02-22 at 4 21 05 PM |  Screen Shot 2023-02-21 at 9 24 29 AM |


aduth (approved):
LGTM 👍 

I'm assuming it was this way previously, but I noticed the "Proofing Over Time" stalled my computer momentarily when generating over about a 3 week period. I also saw a few 403s in the network tab, which I wonder are rate limiting due to requesting too many S3 files over too short a period?

zachmargolis:
> I'm assuming it was this way previously, but I noticed the "Proofing Over Time" stalled my computer momentarily when generating over about a 3 week period. I also saw a few 403s in the network tab, which I wonder are rate limiting due to requesting too many S3 files over too short a period?

I noticed it was slow too, I'll take a closer look

zachmargolis:
This was not approved, I'm going to close out.

PR #88, opened 2023-03-03
Bump dependencies to latest versions

aduth:
Bumps dependencies to latest versions, as well as bumping `.nvmrc` to the latest LTS (18).

Relevant notes:

- `@testing-library/preact-hooks` is undermaintained and its peer dependency with `@testing-library/preact` is invalid, so I removed it here
   - May be worth migrating to React at some point, at least for consistency/ecosystem maintainenance sake
- There's a warning emitted when running tests locally, though this appears to be harmless.
   - >`"import.meta" is not available with the "cjs" output format and will be empty`
- Added `skipLibCheck` to TypeScript configuration to resolve an issue with Preact types (https://github.com/preactjs/preact/issues/3808)
- New Preact types for `JSX.HTMLAttributes` include `SymbolLike` for properties like `href` that conflict with how we were expecting to be able to treat them as strings


zachmargolis (approved):
LGTM

PR #89, opened 2023-03-03
LG-8989: Add "Account Deletions Report"

aduth:
## Screenshot

localhost_5173_account-deletions-report__start=2023-01-29 finish=2023-02-25 env=local


src/hooks/use-registration-data.ts
```diff
@@ -0,0 +1,18 @@
...
+interface RegistrationDataOptions {
+  start: Date;
+  finish: Date;
+}
+
+function useRegistrationData({ start, finish }: RegistrationDataOptions) {
+  const { env } = useContext(ReportFilterContext);
+  const { data } = useQuery(`daily-registrations-${finish.valueOf()}`, () => loadData(finish, env));
+
+  return data?.filter((entry) => entry.date >= start && entry.date <= finish);
```
aduth, line 15: @zachmargolis Wondering if this might explain the performance issues noticed in #87 ? I noticed that we were feeding the full report's set of data to `Plot`, which dates back several years.

zachmargolis: ummm that definitely checks out

src/components/account-deletions-report.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return Plot.plot({
+    color: {
+      type: "ordinal",
+      scheme: "Tableau10",
+    },
+    marks: [Plot.ruleY([0]), Plot.line(data, { x: "date", y: "value" })],
+    width,
+    y: {
+      domain: [0, 0.1],
+    },
```
zachmargolis, line 29: let's add in `tickFormat` here and pass in something to format these as percentages?
```js
import { format  } from 'd3-format'
```

```suggestion
    y: {
      domain: [0, 0.1],
      tickFormat: format(".2%"),
    },
```

(I would move this to a constant and put it in `formats.ts` but the gist is this)

src/components/account-deletions-report.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const width = useElementWidth(ref);
+
+  const formattedData = formatData(data || []);
+
+  console.log(formattedData);
+
+  return (
+    <div ref={ref}>
+      <PlotComponent plotter={() => plot({ data: formattedData, width })} inputs={[data, width]} />
+      <Table numberFormatter={format(".2%")} data={tabulate(formattedData)} />
```
zachmargolis, line 79: yeah I'd extract this format into a constant and share with the `tickFormat`

src/hooks/use-element-width.ts
```diff
@@ -0,0 +1,11 @@
...
+import type { Ref } from "preact/hooks";
+import useResizeListener from "./resize-listener";
+
+function useElementWidth(ref: Ref<HTMLElement>) {
+  const [width, setWidth] = useState(undefined as number | undefined);
+  useResizeListener(() => setWidth(ref.current?.offsetWidth));
+  return width;
+}
+
+export default useElementWidth;
```
zachmargolis, line 11: is there a way to combine this and `useResizeListener` into one hook? or does that not make sense with "the react/preact way"™?

zachmargolis (approved):

src/components/account-deletions-report.tsx
```diff
@@ -0,0 +1,84 @@
...
+  value: number;
+}
+
+function plot({ data, width }: { data: ProcessedFormattedData[]; width?: number }): HTMLElement {
+  return Plot.plot({
+    color: {
+      type: "ordinal",
+      scheme: "Tableau10",
+    },
+    marks: [Plot.ruleY([0]), Plot.line(data, { x: "date", y: "value" })],
```
zachmargolis, line 25: also unsure if it would be useful to have running average like we do in the Proofing Over Time report? With the dotted line?

https://github.com/18F/identity-reporting/blob/6ffea8ddee40ae9223f9fb975f5b089ad59278d7/src/components/proofing-over-time-report.tsx#L345-L379

src/components/account-deletions-report.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function tabulate(results: ProcessedFormattedData[]): TableData {
+  return {
+    header: ["Date", "Percent"],
+    body: results
+      .sort(({ date: aDate }, { date: bDate }) => ascending(aDate, bDate))
+      .map(({ date, value }) => [yearMonthDayFormat(date), value]),
+  };
+}
```
zachmargolis, line 40: the `tabulate` functions lend really well to unit testing, I'd recommend adding tests for this one if possible

src/components/account-deletions-report.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+function getStartOfWeek(date: Date) {
+  const day = date.getDay();
+  const offset = date.getDate() - day;
+  const startOfWeek = new Date(date);
+  startOfWeek.setDate(offset);
+  return startOfWeek;
+}
```
zachmargolis, line 48: we can use a d3 date helper for this (and make sure it works in UTC too), d3.interval.floor  (so `d3.utcWeek.floor` here)

https://github.com/18F/identity-reporting/blob/d78825816f3193775a8fb8ca3cdcc9c1ff1d7c78/src/routes/report-route.tsx#L83

and if we wanted the week to start on Monday there's also [`utcMonday`](https://github.com/d3/d3-time#timeMonday) too etc

src/components/account-deletions-report.tsx
aduth: Yep, will plan to add specs 👍

src/components/account-deletions-report.tsx
aduth: I'll give it a shot 👍

src/hooks/use-element-width.ts
aduth: Yeah, we could. I think having them separate may make sense if they were understandably composable / differentiable and we may plan to use each independently, but in this instance I think we really only care about width. Part of the reason I didn't flatten is because there's a bit of legacy to refactor. I'll see about doing that up-front or as a quick follow-on.

src/components/account-deletions-report.tsx
aduth: Aha, I was wondering how to make the percentages a little nicer looking in the plot.

src/components/account-deletions-report.tsx
aduth: Super helpful, thanks 👍

src/hooks/use-registration-data.ts
zachmargolis: so looking at this, this has the side effect of limiting the data rendered in the Daily Registrations Report (previously that report used to show all-time in the graph, but limit the data in the table). Was that an intentional change here?

aduth: > so looking at this, this has the side effect of limiting the data rendered in the Daily Registrations Report (previously that report used to show all-time in the graph, but limit the data in the table). Was that an intentional change here?

It wasn't intentional, no. I might just make the date range optional so then the daily registrations report can still receive the full data set.

aduth: > > so looking at this, this has the side effect of limiting the data rendered in the Daily Registrations Report (previously that report used to show all-time in the graph, but limit the data in the table). Was that an intentional change here?
> 
> It wasn't intentional, no. I might just make the date range optional so then the daily registrations report can still receive the full data set.

Fixed in afb19f5.

src/components/account-deletions-report.tsx
aduth: > Aha, I was wondering how to make the percentages a little nicer looking in the plot.

Added in cd99a88.

src/components/account-deletions-report.tsx
aduth: Added new format helper in 248da17. Note that I didn't use this for `tickFormat` since it looked better as a round number.

src/hooks/use-element-width.ts
aduth: > is there a way to combine this and `useResizeListener` into one hook? or does that not make sense with "the react/preact way"™?

Flattened in d8aab7d.

src/components/account-deletions-report.tsx
aduth: > also unsure if it would be useful to have running average like we do in the Proofing Over Time report? With the dotted line?

Added running average line in 101af38.

src/components/account-deletions-report.tsx
aduth: > the `tabulate` functions lend really well to unit testing, I'd recommend adding tests for this one if possible

Added tests in 19dd7e5.

src/components/account-deletions-report.tsx
aduth: > we can use a d3 date helper for this (and make sure it works in UTC too), d3.interval.floor (so `d3.utcWeek.floor` here)

Switched to `d3-time` helper in 36e1ac4.

zachmargolis (approved):
LGTM!

src/components/account-deletions-report.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+export function tabulate(results: ProcessedFormattedData[]): TableData {
+  return {
+    header: ["Week Start", "Deleted Users", "Fully Registered Users", "Deletion Rate"],
+    body: results
+      .sort(({ date: aDate }, { date: bDate }) => ascending(aDate, bDate))
+      .map(({ date, deletedUsers, fullyRegisteredUsers, rate }) => [
+        yearMonthDayFormat(date),
+        formatWithCommas(deletedUsers),
+        formatWithCommas(fullyRegisteredUsers),
```
zachmargolis, line 84: I would consider passing the raw number so the CSV doesn't contain commas:
```suggestion
        <td data-csv={deletedUsers}>{formatWithCommas(deletedUsers)}</td>,
        <td data-csv={fullyRegisteredUsers}>{formatWithCommas(fullyRegisteredUsers)}</td>,
```

src/components/account-deletions-report.tsx
```diff
@@ -0,0 +1,124 @@
...
+        formatAsDecimalPercent(rate),
```
zachmargolis, line 85: potentially same here, just pass the decimal to the CSV
```suggestion
        <td data-csv={rate}>{formatAsDecimalPercent(rate)</td>,
```

src/components/account-deletions-report.tsx
aduth: Mm good point, I wondered why `numberFormatter` was exposed the way it was, I guess that explains it.

aduth: Updated in ec47bb7.

src/components/account-deletions-report.tsx
aduth: Updated in ec47bb7.

src/components/account-deletions-report.tsx
```diff
@@ -0,0 +1,124 @@
...
+export function tabulate(results: ProcessedFormattedData[]): TableData {
+  return {
+    header: ["Week Start", "Deleted Users", "Fully Registered Users", "Deletion Rate"],
+    body: results
+      .sort(({ date: aDate }, { date: bDate }) => ascending(aDate, bDate))
+      .map(({ date, deletedUsers, fullyRegisteredUsers, rate }) => [
+        yearMonthDayFormat(date),
+        <td data-csv={deletedUsers}>{formatWithCommas(deletedUsers)}</td>,
+        <td data-csv={fullyRegisteredUsers}>{formatWithCommas(fullyRegisteredUsers)}</td>,
+        <td data-csv={rate}>{formatAsDecimalPercent(rate)}</td>,
```
zachmargolis, line 85: one more suggestion for a few classes to help with rendering (these are normally added for the plain-number table, but for the mix-n-match ones, we gotta do them manually... definitely room to clean these up somehow, maybe a `<TdWithCommas>` `<Td.Percent>` or something component in the future

```suggestion
        <td data-csv={deletedUsers} className="table-number text-tabular text-right">{formatWithCommas(deletedUsers)}</td>,
        <td data-csv={fullyRegisteredUsers} className="table-number text-tabular text-right">{formatWithCommas(fullyRegisteredUsers)}</td>,
        <td data-csv={rate} className="table-number text-tabular text-right">{formatAsDecimalPercent(rate)}</td>,
```

aduth: > definitely room to clean these up somehow, maybe a `<TdWithCommas>` `<Td.Percent>` or something component in the future

Yeah I like that idea 👍

aduth: Added classes in 77b8bcd.

zachmargolis: Gave the components a shot in https://github.com/18F/identity-reporting/pull/93

PR #90, opened 2023-03-06
Fix full path helper for preview environments

aduth:
BASE_URL in preview environments does not include a trailing slash, so avoid trimming the leading slash from the given path.

**Testing Instructions:**

1. Go to https://federalist-14721e0a-0772-4dcd-b62f-ee9d20dcc5fd.sites.pages.cloud.gov/preview/18f/identity-reporting/aduth-fix-preview-links/
2. Click a link in the navigation
3. Refresh the page

Before: 404
After: Current page is shown after refresh


zachmargolis (approved):
LGTM

